PR #1, opened 2020-04-06
first code commit

mhrivnak:
First pass. Obviously there is more to do, such as tests, productization, documentation, etc. but this should be a reasonable point to get feedback and then merge a starting point that works.


pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
```diff
@@ -0,0 +1,79 @@
...
+	// +patchStrategy=merge
+	// +optional
+	Conditions []conditionsv1.Condition `json:"conditions,omitempty"  patchStrategy:"merge" patchMergeKey:"type"`
+}
+
+// Condition Types
+const (
+	// ConditionResourcesExist reports whether all required resources have been created
+	// in the cluster and reflect the specified state.
+	ConditionResourcesExist conditionsv1.ConditionType = "ResourcesExist"
```
mhrivnak, line 52: I dislike the name of this, but I wasn't feeling creative enough to come up with something better. Suggestions are encouraged. This condition really just means that reconciliation completed without error.

pkg/controller/cincinnati/new.go
```diff
@@ -0,0 +1,398 @@
...
+			},
+			Selector: map[string]string{
+				"deployment": nameDeployment(instance),
+			},
+			SessionAffinity: corev1.ServiceAffinityNone,
+		},
+	}
+}
+
+// newPodForCR returns a busybox pod with the same name/namespace as the cr
```
rthallisey, line 139: Left over comment

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
pawanpinjarkar: How about `ConditionClusterResourcesReady` instead of `ConditionResourcesExist`?

mhrivnak: > How about `ConditionClusterResourcesReady` instead of `ConditionResourcesExist`?

This condition only signifies that the operator did all of its work. For example, this operator ensured that a Deployment exists. But, it does not guarantee that the Deployment itself is "ready".

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -0,0 +1,393 @@
...
+		if containers[i].Resources.Requests == nil {
+			containers[i].Resources.Requests = corev1.ResourceList{}
+		}
+		for k, origVal := range original.Resources.Requests {
+			value, ok := containers[i].Resources.Requests[k]
+			if !ok || !origVal.Equal(value) {
+				containers[i].Resources.Requests[k] = origVal
+			}
+		}
+		containers[i].VolumeMounts = original.VolumeMounts
```
djzager, line 239: Some reason these aren't up with the others like it?

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	if err := r.ensureService(ctx, reqLogger, service); err != nil {
+		handleErr(reqLogger, &instance.Status, "EnsureServiceFailed", err)
+		return err
+	}
+	return nil
+}
+
+func (r *ReconcileCincinnati) ensurePEService(ctx context.Context, reqLogger logr.Logger, instance *cv1alpha1.Cincinnati) error {
```
djzager, line 335:
```suggestion
func (r *ReconcileCincinnati) ensurePolicyEngineService(ctx context.Context, reqLogger logr.Logger, instance *cv1alpha1.Cincinnati) error {
```

pkg/controller/cincinnati/new.go
```diff
@@ -0,0 +1,398 @@
...
+			"pe.log.verbosity":               "vv",
+			"pe.mandatory_client_parameters": "channel",
+			"pe.rust_backtrace":              "0",
+			"pe.status.address":              "0.0.0.0",
+			"pe.upstream":                    "http://localhost:8080/v1/graph",
+		},
+	}
+}
+
+func newConfig(instance *cv1alpha1.Cincinnati) (*corev1.ConfigMap, error) {
```
djzager, line 158:
```suggestion
func newGraphBuilderConfig(instance *cv1alpha1.Cincinnati) (*corev1.ConfigMap, error) {
```

pkg/controller/cincinnati/new.go
```diff
@@ -0,0 +1,398 @@
...
+					Path:   "/readiness",
+					Port:   intstr.FromInt(9080),
+					Scheme: corev1.URISchemeHTTP,
+				},
+			},
+		},
+	}
+}
+
+func newPEContainer(instance *cv1alpha1.Cincinnati, image string) corev1.Container {
```
djzager, line 308:
```suggestion
func newPolicyEngineContainer(instance *cv1alpha1.Cincinnati, image string) corev1.Container {
```

tools.go
```diff
@@ -0,0 +1,5 @@
+// +build tools
```
djzager, line 1: Do we want to leave this around _in case_ we need it later?

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
djzager: Yeah I used [`ReconcileCompleted`](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/f8f8f2c6ba1b0eab5e745faf917b1351425b5231/pkg/controller/hyperconverged/hyperconverged_controller.go#L61) to signify that we made it through and did all we wanted to. Not sure if it is what you want.

README.md
```diff
@@ -0,0 +1,10 @@
+# cincinnati-operator
+
```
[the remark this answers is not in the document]
mhrivnak: https://github.com/openshift/cincinnati-operator/issues/3

cmd/manager/main.go
```diff
@@ -0,0 +1,224 @@
...
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"os"
+	"runtime"
+	"strings"
+
+	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
```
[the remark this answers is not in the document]
mhrivnak: I have no idea. This was part of the scaffolding from operator-sdk.

cmd/manager/main.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	// Add support for MultiNamespace set in WATCH_NAMESPACE (e.g ns1,ns2)
+	// Note that this is not intended to be used for excluding namespaces, this is better done via a Predicate
+	// Also note that you may face performance issues when using this with a high number of namespaces.
+	// More Info: https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/cache#MultiNamespacedCacheBuilder
+	if strings.Contains(namespace, ",") {
+		options.Namespace = ""
+		options.NewCache = cache.MultiNamespacedCacheBuilder(strings.Split(namespace, ","))
+	}
```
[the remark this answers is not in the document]
mhrivnak: You guessed right. I don't anticipate needing it, but it seems harmless so I'm inclined to leave it for now.

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -0,0 +1,393 @@
...
+	// appropriate.
+	if err == nil {
+		conditionsv1.SetStatusCondition(&instanceCopy.Status.Conditions, conditionsv1.Condition{
+			Type:    cv1alpha1.ConditionResourcesExist,
+			Status:  corev1.ConditionTrue,
+			Reason:  "Success",
+			Message: "",
+		})
+	}
+	if !reflect.DeepEqual(instanceCopy.Status, instance.Status) {
```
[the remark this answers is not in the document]
mhrivnak: Good point. Is the intent to write out a new heartbeat value every time Reconcile runs? Or just when the condition changes? Seems advantageous, from an api service load perspective, to reduce unnecessary writes.

pkg/controller/cincinnati/cincinnati_controller.go
mhrivnak: I think only because I went in the order that things were specified in the godocs. The `Resources` part above is part of the same flow of setting attributes, but required special care.

tools.go
mhrivnak: I'm not actually sure how we would use this. Do you know?

pkg/controller/cincinnati/cincinnati_controller.go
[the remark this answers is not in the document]
djzager: The intent is to write out the heartbeat every time the condition is computed. That way, there is information to glean from a Condition whose last heartbeat was hours/days ago.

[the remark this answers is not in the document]
mhrivnak: Gotcha. I removed the comparison and am now updating every time.

mhrivnak:
@djzager I think I addressed all your feedback. Let me know if you have anything else.

PR #5, opened 2020-04-08
Ensures Pods get restarted when config changes.

### LINKED ISSUE #2: restart Pods on configmap changes
If one of the ConfigMaps gets changed, we need to ensure those changes are consumed by graph-builder and policy-engine. They may or may not notice changes to the configmap that's mounted on the filesystem. But the other configmap is consumed with the `envFrom` feature, which definitely requires a pod restart.

mhrivnak:
fixes #2

This patch includes two related changes:

1. The Pod has two new annotations that contain a checksum of each ConfigMap.
That way when the contents of a ConfigMap change, that will result in a change
to the Pod template, which will in turn cause the Deployment to replace any
current Pods with a new rollout.

2. The creation of kube resource objects is refactored to happen on a single
object. At the beginning of reconciliation, all resources get created once
up-front and are stored in that object for use throughout the reconciliation
workflow. This is especially helpful since resource creation involves rendering
at least one template, which can return an error; handling that case once
up-front makes it possible to then easily reference resources and their
attributes from other code. Specific to this change, it became necessary to
reference the ConfigMaps from multiple places. Rather than re-render templates
and re-handle error scenarios, it made sense to create the ConfigMaps once and
reference them from a source of truth.

### DIFF
--- a/pkg/controller/cincinnati/cincinnati_controller.go
+++ b/pkg/controller/cincinnati/cincinnati_controller.go
@@ -115,7 +115,14 @@ func (r *ReconcileCincinnati) Reconcile(request reconcile.Request) (reconcile.Re
 	instanceCopy := instance.DeepCopy()
 	instanceCopy.Status = cv1alpha1.CincinnatiStatus{}
 
-	for _, f := range []func(context.Context, logr.Logger, *cv1alpha1.Cincinnati) error{
+	// this object creates all the kube resources we need and then holds them as
+	// the canonical reference for those resources during reconciliation.
+	resources, err := newKubeResources(instanceCopy, r.operandImage)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	for _, f := range []func(context.Context, logr.Logger, *cv1alpha1.Cincinnati, *kubeResources) error{
 		r.ensureConfig,
 		r.ensureEnvConfig,
 		r.ensureDeployment,
@@ -123,7 +130,7 @@ func (r *ReconcileCincinnati) Reconcile(request reconcile.Request) (reconcile.Re
 		r.ensurePolicyEngineService,
 		r.ensurePodDisruptionBudget,
 	} {
-		err = f(ctx, reqLogger, instanceCopy)
+		err = f(ctx, reqLogger, instanceCopy, resources)
 		if err != nil {
 			break
 		}
@@ -160,8 +167,8 @@ func handleErr(reqLogger logr.Logger, status *cv1alpha1.CincinnatiStatus, reason
 	reqLogger.Error(e, reason)
 }
 
-func (r *ReconcileCincinnati) ensureDeployment(ctx context.Context, reqLogger logr.Logger, instance *cv1alpha1.Cincinnati) error {
-	deployment := newDeployment(instance, r.operandImage)
+func (r *ReconcileCincinnati) ensureDeployment(ctx context.Context, reqLogger logr.Logger, instance *cv1alpha1.Cincinnati, resources *kubeResources) error {
+	deployment := resources.deployment
 	if err := controllerutil.SetControllerReference(instance, deployment, r.scheme); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -188,21 +195,30 @@ func (r *ReconcileCincinnati) ensureDeployment(ctx context.Context, reqLogger lo
 	updated.Spec.Replicas = deployment.Spec.Replicas
 	updated.Spec.Selector = deployment.Spec.Selector
 	updated.Spec.Strategy = deployment.Spec.Strategy
+
+	// apply labels an annotations
```
rthallisey, line 199: typo: s/an/and/

rthallisey:
I'm going to end up using some of this code to handle ca cert rotations

djzager:
@mhrivnak My idea was to update:

```
import (
        "crypto/sha256"
)

func nameConfig(instance *cv1alpha1.Cincinnati) string {
	sha := sha256.Sum256([]byte(instance.Spec.Registry + instance.Spec.Repository + instance.Spec.GitHubOrg + instance.Spec.GitHubRepo + instance.Spec.Branch))
	shaString := fmt.Sprintf("%x", sha)
	return instance.Name + "-config" + fmt.Sprintf("%s", shaString[:8]))
}
```

This way, a change to the relevant portion of the Spec would change the name of the configmap and would cause an update to the Deployment.

The only thing that would remain would be for us to clean up old configmaps.

mhrivnak:
I like the simplicity of that idea, but I think it still makes sense to use the annotation.

We also need to restart the Pods if the config changes due to a change in the template or rendering process, even when input to it does not change, so I think it's still important to hash the entire `Data` structure. We could achieve mostly the same effect by adding the template itself as an additional input to the hash. But that could omit edge cases like defaults that change, etc. One other downside is that this would become a bit of code that future devs would have to remember to update when they add or change fields on the CRD's Spec.

Assuming we continue hashing the full Data field, then we'll continue to have an interdependence between resources (which I think is not too uncommon anyway and was probably inevitable), and the idea of creating everything up-front remains important. Then changing the name of the configmap vs. annotating the Pod are still the choices.

It seems like the code and effort it would take to clean up the old configmap after waiting for any related Pods to be deleted could be significant. That would also break the ability for a Deployment to roll back, since the previous Pod definition would reference a ConfigMap that's been deleted. I don't think we have much interest in rollback from the operator's view, but I can see value in supporting this use case: Something breaks, it looks like the operator messed up an update, user turns off the operator and tries to just roll back. But now the configmap has the wrong name, and they don't know why.

This reminds me of the UX "Principle of Least Surprise", which basically says: try not to surprise the user, and given two otherwise equal choices, pick the one that will surprise the user the least. In this case, I think that having "immutable" configmaps that come and go with different names (especially names that appear to have the "random" suffix we see so many other places in k8s) would be more surprising and harder to understand than having a stable long-lived configmap and a weird annotation on the Pod.

WDYT?

djzager:
> This reminds me of the UX "Principle of Least Surprise", which basically says: try not to surprise the user, and given two otherwise equal choices, pick the one that will surprise the user the least. In this case, I think that having "immutable" configmaps that come and go with different names (especially names that appear to have the "random" suffix we see so many other places in k8s) would be more surprising and harder to understand than having a stable long-lived configmap and a weird annotation on the Pod.

I do not believe the idea of immutable configmaps is particularly surprising. There is even [an enhancement](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20191117-immutable-secrets-configmaps.md#proposal) that landed in [k8s 1.18 to add `Immutable` field to ConfigMaps and Secrets](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.18.md#new-api-fields).

> It seems like the code and effort it would take to clean up the old configmap after waiting for any related Pods to be deleted could be significant. That would also break the ability for a Deployment to roll back, since the previous Pod definition would reference a ConfigMap that's been deleted. I don't think we have much interest in rollback from the operator's view, but I can see value in supporting this use case: Something breaks, it looks like the operator messed up an update, user turns off the operator and tries to just roll back. But now the configmap has the wrong name, and they don't know why.

I don't doubt that this could be significant. I do doubt whether it is necessary to solve now. Whenever it would be solved, I would expect you would leave around some n - x (maybe configurable) old configmaps. Either way, how is this scenario solved by the annotation? There would be no rolling back the deployment if something went wrong since there would be nothing to go back to. Assuming we kept around some number of old configmaps (or all until a need arose to do cleanup), using immutable configmaps at least gives you a chance.

> the idea of creating everything up-front remains important.

The only reason I see to create everything up front is to fail early in there are problems rendering the template. I was of the opinion that most of the benefit to having `kuberesources` was being able to call `checksumMap(k.graphBuilderConfig.Data)`.

pkg/controller/cincinnati/new.go
```diff
@@ -388,3 +450,24 @@ func newCMEnvVar(name, key, cmName string) corev1.EnvVar {
...
+		keys = append(keys, k)
+	}
+	keys.Sort()
+
+	hash := sha256.New()
+	encoder := base64.NewEncoder(base64.StdEncoding, hash)
+
+	for _, k := range keys {
+		encoder.Write([]byte(k + ":"))
+		encoder.Write([]byte(m[k] + ":"))
```
sacharya, line 468: Any particular reason for appending  `:` at the end there?

mhrivnak: Generally I'm just trying to use a separator in a simple way that will definitely be distinct from the data. That way `{"foo": "bar"}` and `{"fo": "obar"}` won't serialize the same. Base64-encoding limits the character set, and then `:` was just an arbitrary choice of separator.

mhrivnak: Though now that I'm looking at it again, I don't think I accomplished what I intended. :) I may revise this...

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -115,15 +115,22 @@ func (r *ReconcileCincinnati) Reconcile(request reconcile.Request) (reconcile.Re
 	instanceCopy := instance.DeepCopy()
 	instanceCopy.Status = cv1alpha1.CincinnatiStatus{}
 
-	for _, f := range []func(context.Context, logr.Logger, *cv1alpha1.Cincinnati) error{
+	// this object creates all the kube resources we need and then holds them as
+	// the canonical reference for those resources during reconciliation.
+	resources, err := newKubeResources(instanceCopy, r.operandImage)
```
djzager, line 120: this feels like the kind of error that would be pretty catastrophic. I think we should log it.

pkg/controller/cincinnati/new.go
```diff
@@ -388,3 +450,28 @@ func newCMEnvVar(name, key, cmName string) corev1.EnvVar {
...
+
+	hash := sha256.New()
+	encoder := base64.NewEncoder(base64.StdEncoding, hash)
+
+	// We base64 encode the data to limit the character set and then use ":" as
+	// a separator.
+	for _, k := range keys {
+		encoder.Write([]byte(k))
+		hash.Write([]byte(":"))
+		encoder.Write([]byte(m[k]))
```
rwsu, line 471: Would a Write error be problematic here?

mhrivnak: Yes. Oops. :innocent: Updated. Thanks for the catch!

PR #7, opened 2020-04-10
WIP: graph data init container

rwsu:
This a draft version, that I'm submitting for feedback.

I think we will need to provide an option to include "github-secondary-metadata-scrape" in graphBuilderTOML, so that a user can choose to either use a github repo or an init container as the source of the graph data.

The init container's image needs to be parameterized.


rwsu:
Moved to https://github.com/openshift/cincinnati-operator/pull/10. Closing.

PR #10, opened 2020-04-15
Option to load graph data from an init container

rwsu:
Adds a graphDataImage property to the CRD.

The init container is created when the property is set. If unset
Cincinnati is configured to fetch the graph data from the specified
github repository.

If both graphDataImage and gitHub repository properties are set,
graphDataImage will take precedent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,5 +12,5 @@ operator-sdk run --local
 ## Using an init container to load graph data
 
 The Cincinnati graph data can also be loaded from an init container.
-[docs/graph-data-init-container.md][(docs/graph-data-init-container.md) 
+[docs/graph-data-init-container.md](docs/graph-data-init-container.md) 
 details how.
--- a/README.md
+++ b/README.md
@@ -11,6 +11,4 @@ operator-sdk run --local
 
 ## Using an init container to load graph data
 
-The Cincinnati graph data can also be loaded from an init container.
-[docs/graph-data-init-container.md](docs/graph-data-init-container.md) 
-details how.
+The Cincinnati graph data can also be [loaded from an init container](docs/graph-data-init-container.md).
--- a/README.md
+++ b/README.md
@@ -8,3 +8,9 @@ To run locally, you must set the operand image as shown below.
 export OPERAND_IMAGE="quay.io/app-sre/cincinnati:2873c6b" 
 operator-sdk run --local
 ```
+
+## Using an init container to load graph data
+
+The Cincinnati graph data can also be loaded from an init container.
+[docs/graph-data-init-container.md][(docs/graph-data-init-container.md) 
+details how.
--- a/deploy/crds/cincinnati.openshift.io_cincinnatis_crd.yaml
+++ b/deploy/crds/cincinnati.openshift.io_cincinnatis_crd.yaml
@@ -43,6 +43,12 @@ spec:
               description: GitHubRepo is the repository to use on GitHub for retrieving
                 additional graph data.
               type: string
+            graphDataImage:
+              description: GraphDataImage is an init container image that contains the
+                Cincinnati graph data and copies it to /tmp/cincinnati/graph-data. If 
+                specified, it takes precedence over GitHubOrg and GitHubRepo and disables
+                retrieval of graph data from GitHub.
+              type: string
             registry:
               description: Registry is the container registry to use, such as "quay.io".
               type: string
@@ -58,9 +64,6 @@ spec:
                 as "openshift-release-dev/ocp-release"
               type: string
           required:
-          - branch
-          - gitHubOrg
-          - gitHubRepo
           - registry
           - replicas
           - repository
--- a/deploy/crds/cincinnati.openshift.io_cincinnatis_crd.yaml
+++ b/deploy/crds/cincinnati.openshift.io_cincinnatis_crd.yaml
@@ -45,7 +45,7 @@ spec:
               type: string
             graphDataImage:
               description: GraphDataImage is an init container image that contains the
-                Cincinnati graph data and copies it to /tmp/cincinnati/graph-data. If 
+                Cincinnati graph data and copies it to /var/cincinnati/graph-data. If 
                 specified, it takes precedence over GitHubOrg and GitHubRepo and disables
                 retrieval of graph data from GitHub.
               type: string
--- a/deploy/crds/cincinnati.openshift.io_cincinnatis_crd.yaml
+++ b/deploy/crds/cincinnati.openshift.io_cincinnatis_crd.yaml
@@ -44,10 +44,10 @@ spec:
                 additional graph data.
               type: string
             graphDataImage:
-              description: GraphDataImage is an init container image that contains the
-                Cincinnati graph data and copies it to /var/cincinnati/graph-data. If 
-                specified, it takes precedence over GitHubOrg and GitHubRepo and disables
-                retrieval of graph data from GitHub.
+              description: GraphDataImage is an init container image that contains
+                the Cincinnati graph data and copies it to /var/cincinnati/graph-data.
+                If specified, it takes precedence over GitHubOrg and GitHubRepo and
+                disables retrieval of graph data from GitHub.
               type: string
             registry:
               description: Registry is the container registry to use, such as "quay.io".
@@ -72,7 +72,7 @@ spec:
           description: CincinnatiStatus defines the observed state of Cincinnati
           properties:
             conditions:
-              description: Conditions describes the state of the Cincinnati resource.
+              description: Conditions describe the state of the Cincinnati resource.
               items:
                 description: Condition represents the state of the operator's reconciliation
                   functionality.
--- a/deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
+++ b/deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
@@ -15,4 +15,4 @@ spec:
   # The second is through an init container specified by graphDataImage.
   # When specified, graphDataImage takes precedence over gitHubOrg, 
   # gitHubRepo, and branch. 
-  graphDataImage: "quay.io/rwsu/cincinnati-graph-data-container:latest"
+  #graphDataImage: "your-registry/your-repo/your-init-container"
--- a/deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
+++ b/deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
@@ -6,6 +6,13 @@ spec:
   replicas: 1
   registry: "quay.io"
   repository: "openshift-release-dev/ocp-release"
+  # There two ways to specify the source of the Cincinnati graph data.
+  # The first is through a github repository specified by these three
+  # parameters.
   gitHubOrg: "openshift"
   gitHubRepo: "cincinnati-graph-data"
   branch: "master"
+  # The second is through an init container specified by graphDataImage.
+  # When specified, graphDataImage takes precedence over gitHubOrg, 
+  # gitHubRepo, and branch. 
+  graphDataImage: "quay.io/rwsu/cincinnati-graph-data-container:latest"
--- a/dev/Dockerfile
+++ b/dev/Dockerfile
@@ -1,0 +1,6 @@
+FROM registry.access.redhat.com/ubi8/ubi:8.1
+
+RUN dnf install -y git
+RUN git clone https://github.com/openshift/cincinnati-graph-data
+
+CMD exec /bin/bash -c "cp -rf /cincinnati-graph-data/* /var/cincinnati/graph-data/"
--- a/docs/graph-data-init-container.md
+++ b/docs/graph-data-init-container.md
@@ -1,0 +1,48 @@
+# Graph Data Init Container
+
+Normally, the graph data is fetched directly from the Cincinnati graph
+data repository at: http://github.com/openshift/cincinnati-graph-data.
+In environments where an internet connection is not available, loading
+from an init container is another way to make the graph data available
+to Cincinnati.
+
+The role of the init container is to provide a copy of the graph data.
+During pod initialization, the init container copies the data to a volume
+at /var/cincinnati/graph-data. The graph builder is configured to read 
+the graph data from the same location.
+
+## Build the graph data init container
+
+An example of how to build an init container can be found in ./dev/Dockerfile.
+In the example, the image clones the Cincinnati graph data repository,
+http://github.com/openshift/cincinnati-graph-data.
+
+````
+podman build -f ./dev/Dockerfile -t quay.io/rwsu/cincinnati-graph-data-
+container:latest
+podman push quay.io/rwsu/cincinnati-graph-data-container:latest
+````
+
+## Configure the operator to use the init container
+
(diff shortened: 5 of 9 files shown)


deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
```diff
@@ -6,6 +6,13 @@ spec:
...
+  # parameters.
   gitHubOrg: "openshift"
   gitHubRepo: "cincinnati-graph-data"
-  branch: "master"
\ No newline at end of file
+  branch: "master"
+  # The second is through an init container specified by graphDataImage.
+  # When specified, graphDataImage takes precedence over gitHubOrg, 
+  # gitHubRepo, and branch. 
+  graphDataImage: "quay.io/rwsu/cincinnati-graph-data-container:latest"
```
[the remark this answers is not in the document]
djzager: It would be nice if we could build from this project (ie. `quay.io/openshift/cincinnati-graph-data-container`)

pkg/controller/cincinnati/new.go
```diff
@@ -62,6 +63,31 @@ data_directory = "/tmp/cincinnati/graph-data"
 [[plugin_settings]]
 name = "edge-add-remove"`
 
+// TOML when graph data source is an init container
+const graphBuilderTOMLInitContainer string = `verbosity = "vvv"
```
djzager, line 67: If we can, I believe we should have one template to rule them all. Tinkering around with this for a few minutes and I came up with https://play.golang.org/p/eHIbv22xAeI WDYT?

pkg/controller/cincinnati/new.go
```diff
@@ -301,6 +344,27 @@ func (k *kubeResources) newDeployment(instance *cv1alpha1.Cincinnati) *appsv1.De
...
+	if k.UseInitContainer(instance) {
+		dep.Spec.Template.Spec.InitContainers = []corev1.Container{
+			*k.graphDataInitContainer,
+		}
+	}
+
+	return dep
+}
+
+func (k *kubeResources) newGraphDataInitContainer(instance *cv1alpha1.Cincinnati, image string) *corev1.Container {
```
djzager, line 356:
```suggestion
func (k *kubeResources) newGraphDataInitContainer(instance *cv1alpha1.Cincinnati) *corev1.Container {
```

You already have the instance.

pkg/controller/cincinnati/new.go
```diff
@@ -107,6 +134,9 @@ func newKubeResources(instance *cv1alpha1.Cincinnati, image string) (*kubeResour
 	k.envConfigHash = envConfigHash
 	k.podDisruptionBudget = k.newPodDisruptionBudget(instance)
 	k.graphBuilderContainer = k.newGraphBuilderContainer(instance, image)
+	if k.UseInitContainer(instance) {
```
djzager, line 137: I think you could update `newGraphDataInitContainer` to simply return `nil` if the `.Spec.GraphDataImage == ""` and update this to be:

```suggestion
	k.graphDataInitContainer = k.newGraphDataInitContainer(instance)
```

pkg/controller/cincinnati/new.go
```diff
@@ -301,6 +344,27 @@ func (k *kubeResources) newDeployment(instance *cv1alpha1.Cincinnati) *appsv1.De
...
+
+func (k *kubeResources) newGraphDataInitContainer(instance *cv1alpha1.Cincinnati, image string) *corev1.Container {
+	return &corev1.Container{
+		Name:            NameInitContainerGraphData,
+		Image:           image,
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		VolumeMounts: []corev1.VolumeMount{
+			corev1.VolumeMount{
+				Name:      "cincinnati-graph-data",
+				MountPath: "/tmp/cincinnati/graph-data",
```
djzager, line 364: I doubt this matters in the case of matters (that /tmp doesn't survive reboot), but I wonder if we should be using a non `tmp` directory for the graph-data?

pkg/controller/cincinnati/new.go
```diff
@@ -496,3 +564,7 @@ func checksumMap(m map[string]string) (string, error) {
 
 	return fmt.Sprintf("%x", hash.Sum(nil)), nil
 }
+
+func (k *kubeResources) UseInitContainer(instance *cv1alpha1.Cincinnati) bool {
```
djzager, line 568: I don't think we need this if `newGraphDataInitContainer` is updated to return `nil` if the image isn't specified.

pkg/controller/cincinnati/new.go
```diff
@@ -62,6 +63,31 @@ data_directory = "/tmp/cincinnati/graph-data"
 [[plugin_settings]]
 name = "edge-add-remove"`
 
+// TOML when graph data source is an init container
```
mhrivnak, line 66: Did you consider using just one template, and putting a conditional around the github part? Perhaps you could have that rendered only if `instance.GraphDataImage != ""`?

pkg/controller/cincinnati/new.go
```diff
@@ -301,6 +344,27 @@ func (k *kubeResources) newDeployment(instance *cv1alpha1.Cincinnati) *appsv1.De
 			},
 		},
 	}
+	if k.UseInitContainer(instance) {
```
djzager, line 347: If you update `k.graphDataInitContainer` to return `nil` when the image isn't specified, then you should be able to include it with the deployment definition. Ie:

```go
				Spec: corev1.PodSpec{
...
					Containers: []corev1.Container{
						*k.graphBuilderContainer,
						*k.policyEngineContainer,
					},
					InitContainers: []corev1.Container{
						*k.graphBuilderContainer,
					},
				},
```

pkg/controller/cincinnati/new.go
djzager: I am in agreement with this idea: https://github.com/openshift/cincinnati-operator/pull/10#discussion_r408837314

pkg/controller/cincinnati/new.go
mhrivnak: Wouldn't that cause a nil pointer dereference?

djzager: downside of not being able to make :coffee: I guess.

deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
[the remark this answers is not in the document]
mhrivnak: For now I think we should avoid any appearance that there might be an "official" build of this init container until we get agreement that as Red Hat we want to produce, ship, and support one. It would be easy for someone to see this feature, see that repo, and just start using it in production because it works. For now this feature is intended for a user who is willing to setup their own build of an init container that snapshots the data at whatever point in time they choose.

I definitely like the idea of having one available for development and testing; I'd just keep it under some less-official-looking namespace. Maybe `quay.io/cincinnati/scarybreakseverythinggraphdata` or similar. :wink: Ok maybe not that, but something clearly unofficial.

pkg/controller/cincinnati/new.go
djzager: Still think we can get around the `UseInitContainer` by letting `graphBuilderContainer` return `nil` and saying:

```go
if k.graphBuilderContainer != nil {
    ...Spec.InitContainers = []corev1.Container{*k.graphBuilderContainer}
}
```

pkg/controller/cincinnati/new.go
mhrivnak: I had the same thought. I'm a fan of using the [FHS](http://refspecs.linuxfoundation.org/fhs.shtml) as intended even in containers. The only reason I didn't bring it up is that we got this value from the [templates in cincinnati's own repo](https://github.com/openshift/cincinnati/blob/c1e2b4e/dist/openshift/cincinnati.yaml#L329). :shrug: I could go either way on it. Since it's a volume formally being mounted in, something in `/var/` could be perfectly reasonable.

pkg/controller/cincinnati/new.go
mhrivnak: Another option would be to just have `k.initContainers = newInitContainers(instance)`, and let it return a slice of whatever length it thinks is appropriate. Then in the Deployment Spec you can always assign the value, and if it's an empty slice, no harm. Just another option, and I don't personally have a preference.

deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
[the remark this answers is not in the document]
djzager: I think a suitable compromise would be:

1. Not including the `graphDataImage` in the basic CR at all
1. Introducing the `Dockerfile` one could use to build such an image next to relevant documentation about how/why to use it. Probably best to create an issue for the documentation and just add the `Dockerfile` now (maybe in `$PROJECT_ROOT/dev)?

pkg/controller/cincinnati/new.go
djzager: It is the VolumeMount that grabs my attention. `/var/` sounds good to me.

pkg/controller/cincinnati/names.go
```diff
@@ -9,6 +9,8 @@ const (
 	NameContainerGraphBuilder string = "graph-builder"
 	// NameContainerPolicyEngine is the Name property of the policy engine container
 	NameContainerPolicyEngine string = "policy-engine"
+	// NameContainerPolicyEngine is the Name property of the policy engine container
```
sacharya, line 12: // NameInitContainerGraphData is the Name property of the graph data container

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
```diff
@@ -33,6 +33,10 @@ type CincinnatiSpec struct {
 	// Branch is the git branch to use on GitHub for retrieving additional graph
 	// data.
 	Branch string `json:"branch"`
+
+	// GraphDataImage is an init container image that contains the Cincinnati
+	// graph data and copies it to /tmp/cincinnati/graph-data.
```
sacharya, line 38: Change to /var/cincinnati/graph-data

deploy/crds/cincinnati.openshift.io_cincinnatis_crd.yaml
```diff
@@ -43,6 +43,12 @@ spec:
               description: GitHubRepo is the repository to use on GitHub for retrieving
                 additional graph data.
               type: string
+            graphDataImage:
+              description: GraphDataImage is an init container image that contains the
+                Cincinnati graph data and copies it to /tmp/cincinnati/graph-data. If
```
sacharya, line 48: Change to /var/cincinnati/graph-data

docs/graph-data-init-container.md
```diff
@@ -0,0 +1,48 @@
+# Graph Data Init Container
+
+Normally, the graph data is fetched directly from the Cincinnati graph
+data repository at: http://github.com/openshift/cincinnati-graph-data.
```
wking, line 4: nit: http -> https  Also might want to make this a formal link so readers don't have to think about Markdown link detection vs. trailing periods.

README.md
```diff
@@ -8,3 +8,9 @@ To run locally, you must set the operand image as shown below.
 export OPERAND_IMAGE="quay.io/app-sre/cincinnati:2873c6b" 
 operator-sdk run --local
 ```
+
+## Using an init container to load graph data
+
+The Cincinnati graph data can also be loaded from an init container.
+[docs/graph-data-init-container.md](docs/graph-data-init-container.md)
```
wking, line 15: nit: No need for the full path on the human side, just:

```markdown
The Cincinnati graph data can also be [loaded from an init container](docs/graph-data-init-container.md).
```

deploy/crds/cincinnati.openshift.io_cincinnatis_crd.yaml
```diff
@@ -43,6 +43,12 @@ spec:
               description: GitHubRepo is the repository to use on GitHub for retrieving
                 additional graph data.
               type: string
+            graphDataImage:
+              description: GraphDataImage is an init container image that contains the
```
djzager, line 47: Did you add this description manually or via comment? I think we should rely on the CRD being generated from controller-tools (so we don't accidentally lose something important in the future).

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -150,6 +150,21 @@ func (r *ReconcileCincinnati) Reconcile(request reconcile.Request) (reconcile.Re
 			Message: "",
 		})
 	}
+	if instance.Spec.GraphDataImage != "" && (instance.Spec.GitHubOrg != "" || instance.Spec.GitHubRepo != "" || instance.Spec.Branch != "") {
+		conditionsv1.SetStatusCondition(&instanceCopy.Status.Conditions, conditionsv1.Condition{
+			Type:    cv1alpha1.ConditionConfigurationConflict,
+			Status:  corev1.ConditionTrue,
+			Reason:  "ConflictFound",
+			Message: "graphDataImage and gitHubOrg/gitHubRepo/branch should not be configured at the same time.",
```
djzager, line 158: Should also mention that the `graphDataImage` is winning.

rwsu: Done.

deploy/crds/cincinnati.openshift.io_cincinnatis_crd.yaml
rwsu: It was manually. Thanks for the reminder. I've updated _types.go and regenerated the CRD using operator-sdk.

deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
```diff
@@ -6,6 +6,13 @@ spec:
   replicas: 1
   registry: "quay.io"
   repository: "openshift-release-dev/ocp-release"
+  # There two ways to specify the source of the Cincinnati graph data.
```
mhrivnak, line 9: s/There/There are/

dev/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM registry.access.redhat.com/ubi8/ubi:8.1
+
+RUN dnf install -y git
+RUN git clone https://github.com/openshift/cincinnati-graph-data
```
mhrivnak, line 4: Improvement idea for later: I think it would be advantageous to download just a [tarball of master HEAD](https://github.com/openshift/cincinnati-graph-data/archive/master.tar.gz). I did it just now, and it was only 16kB. It's also just one file to deal with instead of all the diffs that get downloaded by git, AND that way you don't even need to install git.

It would then just be a good idea to record somewhere in the image what commit hash was downloaded. Might also be a good label on the image.

But I think this is fine for now.

dev/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+
+CMD exec /bin/bash -c "cp -rf /cincinnati-graph-data/* /var/cincinnati/graph-data/"
```
mhrivnak, line 6: nit, fixable later if you want: suggest `/var/lib/` instead of just `/var/`, which I think best-matches the FHS.

PR #11, opened 2020-04-15
Support External Registry CA certs

rthallisey:
OpenShift uses the image.config.openshift.io API to store external registry information so it can be used in the cluster.  This PR will look at the field 'additionalTrustedCA' on the image.config.openshift.io singleton to get the ConfigMap name that holds the external registry's cert.  The operator will make a copy of this configmap in the local namespace and mount it into the Cincinnati deployment.  Changes in the configmap will results in Cincinnati being restarted, so certs this also supports cert rotation.


rthallisey:
I'm not certain which config map, the local or the source configmap, is being watched.  Anyone know?

rthallisey:
@mhrivnak @djzager all set for review

deploy/operator.yaml
```diff
@@ -15,15 +15,13 @@ spec:
       serviceAccountName: cincinnati-operator
       containers:
         - name: cincinnati-operator
-          image: quay.io/cincinnati/cincinnati-operator:latest
+          image: local.registry:5000/rhallise/cincinnati-operator
```
djzager, line 18: should this change be committed?

deploy/role.yaml
```diff
@@ -90,3 +90,104 @@ rules:
...
+  name: cincinnati-operator
+rules:
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - images
+  verbs:
+  - get
+  - list
+  - watch
```
djzager, line 107: I don't know if we care to limit it but I think you can also limit it to a particular resourceName.

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
```diff
@@ -39,6 +39,12 @@ type CincinnatiSpec struct {
 	// specified, it takes precedence over GitHubOrg and GitHubRepo and disables
 	// retrieval of graph data from GitHub.
 	GraphDataImage string `json:"graphDataImage,omitempty"`
+
+	// CmKey refers to the ConfigMap key that holds a CA cert to the registry
+	// Cincinnati will contact to build the upgrade graph.  The name of the
+	// ConfigMap holding the key comes from the field additionalTrustedCA
+	// in the image.config.openshift.io API.
+	CmKey string `json:"cmKey"`
```
djzager, line 47: either `ConfigmapKey` seems appropriate.

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -183,6 +188,72 @@ func handleErr(reqLogger logr.Logger, status *cv1alpha1.CincinnatiStatus, reason
 	reqLogger.Error(e, reason)
 }
 
+func (r *ReconcileCincinnati) ensureAdditionalTrustedCA(ctx context.Context, reqLogger logr.Logger, instance *cv1alpha1.Cincinnati, resources *kubeResources) error {
+	defaultOpenShiftConfigNS := "openshift-config"
+
+	// Check if the Cluster is aware of a registry requiring an
+	// AdditionalTrustedCA
+	regconf := &apicfgv1.RegistrySources{}
+	image := &apicfgv1.Image{
```
djzager, line 197: `import "k8s.io/apimachinery/pkg/types"`
```suggestion
	image := &apicfgv1.Image{}
	err = r.client.Get(ctx, types.NamespacedName{Name: defaults.ImageConfigName}, image)
```

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
djzager: Don't forget to regenerate the CRD.

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -183,6 +188,72 @@ func handleErr(reqLogger logr.Logger, status *cv1alpha1.CincinnatiStatus, reason
 	reqLogger.Error(e, reason)
 }
 
+func (r *ReconcileCincinnati) ensureAdditionalTrustedCA(ctx context.Context, reqLogger logr.Logger, instance *cv1alpha1.Cincinnati, resources *kubeResources) error {
+	defaultOpenShiftConfigNS := "openshift-config"
```
djzager, line 192: I think this should be a package level `const` and just `openshiftConfigNamespace`

pkg/controller/cincinnati/cincinnati_controller.go
djzager: Maybe this didn't work for you before? but `ObjectKeyFromObject` returns an [ObjectKey](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#ObjectKey) which is a NamespacedName.

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -183,6 +188,72 @@ func handleErr(reqLogger logr.Logger, status *cv1alpha1.CincinnatiStatus, reason
...
+	err = r.client.Get(ctx, types.NamespacedName{Name: image.Spec.AdditionalTrustedCA.Name, Namespace: defaultOpenShiftConfigNS}, sourceCM)
+	if err != nil && errors.IsNotFound(err) {
+		reqLogger.Info("Found image.config.openshift.io.Spec.AdditionalTrustedCA.Name but did not find expected ConfigMap", "Name", image.Spec.AdditionalTrustedCA.Name, "Namespace", defaultOpenShiftConfigNS)
+		return err
+	} else if err != nil {
+		return err
+	}
+
+	localCM := &corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
```
djzager, line 235: is there a reason you can't skip the TypeMeta?

pkg/controller/cincinnati/cincinnati_controller.go
djzager: Should we have a watch on these?

pkg/controller/cincinnati/new.go
```diff
@@ -318,6 +316,26 @@ func (k *kubeResources) newDeployment(instance *cv1alpha1.Cincinnati) *appsv1.De
 		}
 	}
 
+	if instance.Spec.CmKey != "" {
+		v := corev1.Volume{
+			Name: "trusted-ca",
```
djzager, line 321: Should make `"trusted-ca"`, at a minimum, a package local variable so we don't mess it up w.r.t. the container's volume mount.

deploy/role.yaml
rthallisey: I'm fine either way.  This is a singleton resource, so we're pretty much limited out of the box.

pkg/controller/cincinnati/cincinnati_controller.go
rthallisey: yes, we mainly need to watch the openshift-config configmap since that's our source

rthallisey: What configmap is being watched, I'm not certain of.  I was wondering that earlier too: https://github.com/openshift/cincinnati-operator/pull/11#issuecomment-614260225

pkg/controller/cincinnati/cincinnati_controller.go
rthallisey: this is some leftover frustration, along with a few other spots.

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
```diff
@@ -39,6 +39,12 @@ type CincinnatiSpec struct {
 	// specified, it takes precedence over GitHubOrg and GitHubRepo and disables
 	// retrieval of graph data from GitHub.
 	GraphDataImage string `json:"graphDataImage,omitempty"`
+
+	// CofigmapKey refers to the ConfigMap key that holds a CA cert to the registry
+	// Cincinnati will contact to build the upgrade graph.  The name of the
+	// ConfigMap holding the key comes from the field additionalTrustedCA
+	// in the image.config.openshift.io API.
+	ConfigmapKey string `json:"configmapKey"`
```
sacharya, line 47: Make it `ConfigMapKey` with capital M? Or even `CertConfigMapKey` to give it some context.

rthallisey: Ya I like CertConfigMapKey better than ConfigmapKey.

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
```diff
@@ -39,6 +39,12 @@ type CincinnatiSpec struct {
 	// specified, it takes precedence over GitHubOrg and GitHubRepo and disables
 	// retrieval of graph data from GitHub.
 	GraphDataImage string `json:"graphDataImage,omitempty"`
+
+	// CertConfigMapKey refers to the ConfigMap key that holds a CA cert to the registry
+	// Cincinnati will contact to build the upgrade graph.  The name of the
+	// ConfigMap holding the key comes from the field additionalTrustedCA
+	// in the image.config.openshift.io API.
+	CertConfigMapKey string `json:"certConfigMapKey"`
```
mhrivnak, line 47: Just confirming: is this not a predictable value? Is there not a way for us to either know it up-front or automatically figure it out?

mhrivnak, line 47: I think this should be optional, particularly for connected use cases.

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -183,6 +187,49 @@ func handleErr(reqLogger logr.Logger, status *cv1alpha1.CincinnatiStatus, reason
 	reqLogger.Error(e, reason)
 }
 
+func (r *ReconcileCincinnati) ensureAdditionalTrustedCA(ctx context.Context, reqLogger logr.Logger, instance *cv1alpha1.Cincinnati, resources *kubeResources) error {
+	// Check if the Cluster is aware of a registry requiring an
+	// AdditionalTrustedCA
+	image := &apicfgv1.Image{}
+	err := r.client.Get(ctx, types.NamespacedName{Name: defaults.ImageConfigName, Namespace: ""}, image)
+	if err != nil {
```
mhrivnak, line 195: Is a "not found" error possible here? Would some clusters just not have this resource?

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
rthallisey: It's not predictable.  See the example from these docs: https://docs.openshift.com/container-platform/4.2/registry/configuring-registry-operator.html#registry-operator-config-resources-configmap_configuring-registry-operator.  What's good is that there will be only one configmap to read from, but there can be multiple keys.  They keys are normally a registry's hostname, so Cincinnati wouldn't know which one to pick.

pkg/controller/cincinnati/cincinnati_controller.go
mhrivnak: I suggested new Watches just now in my review comment below.

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
mhrivnak: Wouldn't it just be the `Registry` field on the Cincinnati Spec?

rthallisey: We can run into danger if we assume `Registry` equals the Image Registry key because the object is a configmap key, so any key is acceptable.  Therefore, I want to avoid equating the two because there is a potential bug here when a user sets Registry=my.local.registry and there happens to already be a key in this configmap with 'my.local.registry'.  If the configmap keys were gated in some way as to assure they were valid registries and working keys, then I would agree.

pkg/controller/cincinnati/cincinnati_controller.go
rthallisey: It's possible, but it gets scooped up in the if block.

My thinking here is that if there's any error, back out and ignore any CA request.  Line 199 expands on this by catching the case where the object exists, but there's no `AdditionalTrustedCA`.  The last check is line 205, which makes sure the configmap with the expected name actually exists.   So after line 205, I would expect to always a valid configmap.    

I also don't check for a valid `CertConfigMapKey` because kubernetes will handle that when the mount fails.

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
rthallisey: Right, it's optional.  If you set a key and your registry doesn't take one you'll have errors, so `""`, or not including it at all, is a perfectly acceptable option.

mhrivnak: The docs say "The key is the host name of a registry with the port for which this CA is to be trusted." That seems clear. I'm not following where the uncertainty is. It's a map of hostnames to CAs, and it seems like we should just look for the hostname we care about to appear in that map. Surely that's what the OCP image registry itself is doing, right?

rthallisey: I'm not sure how OCP image registry uses this configmap.  My expectation is that the image registry uses key/value to allow you to reach external registry images through the internal registry address.

What were weighing here is the assumption that our valid registry string is equal to an unverified registry string, versus the user typing in a string.  Bottom line, we're using this API as if it were a configmap, so I'm not too thrilled about the idea of assuming key equivalence when the trade-off is small. 

Before landing on using this API, I was going to use a configmap we owned (we talked about this already).   Going with the API seemed like a better fit only because this looked like the general place folks would store this data.  That being said, I'd be fine with removing this field if we owned the data transmission format and could guarantee the relationship. 

Regardless, I think both sides of this is splitting hairs because we're talking about a small UX gain against major user error.  It's unusual, but I guess that makes me the user pessimist here.  Maybe the OTA team can decide the fate of this field, preferably outside of this PR. WDYT?

mhrivnak: I guess I'm thinking that >99% of the time the values are going to be the same. I guess we could later remove this field if we find that they're the same 100% of the time.

Do you have an example scenario in mind that might help me understand the concern better?

pkg/controller/cincinnati/cincinnati_controller.go
mhrivnak: If there's a connected cluster where this resource just doesn't exist at all, and isn't needed at all, seems like this function should not return an error. Returning an error stops reconciliation and sets an error condition. WDYT?

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
mhrivnak: Ah, I was thinking it should have `omitempty`, but perhaps defaulting to empty string is enough.

rthallisey: Admin (a different ryanh) wants a place to store his external registry cert, so he creates an entry in the image config API with `key: ryanh`, that isn't a resolvable name from inside the pods.  Then, the admin gives me access to his cluster.  I happen to run my own registry located at `ryanh` and I make it so `ryanh` will resolve to my registry in the cluster.  If I go to run Cincinnati at this point, a number of things can go wrong depending what the Admin uses that cert for, whether I followed the docs or not, ect....

rthallisey: My thinking here is that we don't have API overload and it's good to raise awareness of this key. Could go either way here.

pkg/controller/cincinnati/cincinnati_controller.go
rthallisey: I see what you're saying.  It should returning `nil`.

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
mhrivnak: I guess I'm not seeing the value in why someone would put a CA in there whose key isn't a valid hostname.

I'm good with merging this as-is, and let's check in next week with the image registry operator folks just to double-check what assumptions we can or can't make about the contents.

rthallisey: > assumptions we can or can't make about the contents.
>
👍

djzager: My vote is for `omitempty`

djzager:
```suggestion
	// +kubebuilder:validation:Optional
	CertConfigMapKey string `json:"certConfigMapKey"`
```

rthallisey: we'll `omitempty` then.

rthallisey:
thanks for the reviews!

rthallisey:
Some open items after this PR:
- Finalize API
- Add watches on new secondary resources

pkg/apis/cincinnati/v1alpha1/cincinnati_types.go
rthallisey: I needed to sleep on this.  The problem with my argument is, while yes it's true that we have to trust unverified keys, that's the same trust you extend to any configmap you consume. 

My perspective on this is driven by the fact that I don't trust we're using this API exactly the way it's intended or something is missing.  This description makes me think it's the right place: `AdditionalTrustedCA in the image.config.openshift.io resource to provide additional CAs that should be trusted when contacting external registries.`  But why would anyone that wants registry certs in a pod go through all this trouble creating a key and copying a configmap locally when they can just create a local configmap?   It just seems like the registry operator is incomplete.  We'll talk about this more next week, have a good weekend.

PR #13, opened 2020-04-16
Adding some unit tests for the controller

sacharya:


pkg/controller/cincinnati/cincinnati_controller_test.go
```diff
@@ -0,0 +1,140 @@
...
+	cincinnati := newDefaultCincinnati()
+	r := newTestReconciler(cincinnati)
+	request := newRequest(cincinnati)
+
+	_, err := r.Reconcile(request)
+	if err != nil {
+		t.Fatal(err)
+	}
+	instance := &cv1alpha1.Cincinnati{}
+	r.client.Get(nil, request.NamespacedName, instance)
```
pawanpinjarkar, line 63: Do not pass a nil Context, even if a function permits it; pass `context.TODO`

pkg/controller/cincinnati/cincinnati_controller_test.go
```diff
@@ -0,0 +1,140 @@
...
+	cincinnati := newDefaultCincinnati()
+	r := newTestReconciler(cincinnati)
+
+	resources, err := newKubeResources(cincinnati, testOperandImage)
+	err = r.ensureConfig(nil, log, cincinnati, resources)
+	if err != nil {
+		t.Fatal(err)
+	}
+	found := &corev1.ConfigMap{}
+	err = r.client.Get(nil, types.NamespacedName{Name: nameConfig(cincinnati), Namespace: cincinnati.Namespace}, found)
```
pawanpinjarkar, line 78: Same as above. Please make the change for the remaining test cases.

pkg/controller/cincinnati/cincinnati_controller_test.go
```diff
@@ -0,0 +1,140 @@
...
+	assert.True(t, conditionsv1.IsStatusConditionTrue(instance.Status.Conditions, cv1alpha1.ConditionReconcileCompleted))
+}
+
+// TestEnsureConfig ensures that Config is created successully and the OwnerReference is set.
+func TestEnsureConfig(t *testing.T) {
+	cincinnati := newDefaultCincinnati()
+	r := newTestReconciler(cincinnati)
+
+	resources, err := newKubeResources(cincinnati, testOperandImage)
+	err = r.ensureConfig(nil, log, cincinnati, resources)
```
pawanpinjarkar, line 73: Same as above.

pawanpinjarkar:
Using `context.TODO` instead of `nil` in test cases safeguards us from build failures (at least) if we happen to implement the same code style/linting pattern as other repo like foundation in the future.

pkg/controller/cincinnati/cincinnati_controller_test.go
sacharya: Sounds like a sensible default.

PR #15, opened 2020-04-16
Use /var/lib and tarball for graph data

rwsu:


dev/Dockerfile
```diff
@@ -1,6 +1,5 @@
 FROM registry.access.redhat.com/ubi8/ubi:8.1
 
-RUN dnf install -y git
-RUN git clone https://github.com/openshift/cincinnati-graph-data
+RUN curl -o cincinnati-graph-data.tar.gz https://codeload.github.com/openshift/cincinnati-graph-data/tar.gz/master
```
sacharya, line 3: Consider following redirect on the actual url instead of using the derived url.
`curl -L -o cincinnati-graph-data.tar.gz https://github.com/openshift/cincinnati-graph-data/archive/master.tar.gz`
It's more likely to not change in the future.

rwsu: Done

PR #19, opened 2020-04-21
Adding more unit tests

sacharya:


pkg/controller/cincinnati/cincinnati_controller_test.go
```diff
@@ -52,17 +57,47 @@ func TestMain(m *testing.M) {
...
+				t.Fatal(err)
+			}
+			instance := &cv1alpha1.Cincinnati{}
+			err = r.client.Get(context.TODO(), request.NamespacedName, instance)
+			if err != nil {
+				t.Fatal(err)
+			}
+			assert.True(t, conditionsv1.IsStatusConditionTrue(instance.Status.Conditions, cv1alpha1.ConditionReconcileCompleted))
+
+			assert.Equal(t, instance.Status.Conditions[1].Type, cv1alpha1.ConditionConfigurationConflict)
```
pawanpinjarkar, line 97: Instead of checking multiple status conditions in different ways, how about using `expectedConditions []conditionsv1.Condition` in the struct definition, set the required `Type` and `Status` in the test case and then loop through it to check for conditions,
```
	for _, condition := range test.expectedConditions {
		assert.True(t, conditionsv1.IsStatusConditionPresentAndEqual(instance.Status.Conditions, condition.Type, condition.Status))
	}
	if instance != nil {
		assert.Equal(t, len(test.expectedConditions), len(test.instance.Status.Conditions))
	}
```
This then can be added in a utility function and other testcases can use it as well.

sacharya: Will do it if there is a bit more repetition. Right now, it's just 2 lines. :smile:

djzager: It's odd that we are relying on `instance.Status.Conditoins[1].Type` to be the `ConditionConfigurationConflict`...which means that we know another condition should be defined that we aren't checking. My belief is that we should be explicit in what we want the conditions to be after calling the function under test (in this case `Reconcile`) and verify that those conditions are there (and no others).

sacharya: Fair enough. Added expectedConditions to the struct.

pkg/controller/cincinnati/cincinnati_controller_test.go
```diff
@@ -100,8 +148,215 @@ func TestEnsureEnvConfig(t *testing.T) {
...
+			cincinnati := newDefaultCincinnati()
+			r := newTestReconciler(test.existingObjs...)
+
+			resources, err := newKubeResources(cincinnati, testOperandImage)
+			err = r.ensureAdditionalTrustedCA(context.TODO(), log, cincinnati, resources)
+
+			verifyError(t, err, test.expectedError)
+
+			if test.name == "ConfigMapCreate" {
+				found := &corev1.ConfigMap{}
```
sacharya, line 209: > I don't see a specific reason why we need 3 versions of TestEnsureAdditionalTrustedCA

In a success scenario, you may want to verify more things like resources are created properly, that you typically may not care about in a failure scenario. Like In this particular case, I ended up with a  if-condition to handle that case. Getting rid of that if-condition would need expectedConfigMap object to be passed to the struct, which means I'd have to construct that object. Just trying to find the right balance here.

pkg/controller/cincinnati/cincinnati_controller_test.go
```diff
@@ -139,3 +398,50 @@ func newDefaultCincinnati() *cv1alpha1.Cincinnati {
...
+func verifyConditions(t *testing.T, expectedConditions []conditionsv1.Condition, cincinnati *cv1alpha1.Cincinnati) {
+	for _, condition := range expectedConditions {
+		assert.True(t, conditionsv1.IsStatusConditionPresentAndEqual(cincinnati.Status.Conditions, condition.Type, condition.Status))
+	}
+	assert.Equal(t, len(expectedConditions), len(cincinnati.Status.Conditions))
+}
+
+func verifyOwnerReference(t *testing.T, ownerReference metav1.OwnerReference, cincinnati *cv1alpha1.Cincinnati) {
+	assert.Equal(t, ownerReference.Name, cincinnati.Name)
+	//Note: These properties seem to be derived from runtime object and do not seem to match the expected values.
```
pawanpinjarkar, line 436: Is this commented code here on purpose? Can it be removed?

sacharya: I'd like to leave it there, in case it trips up someone else too. I need to dig into it further and may be file a bug.

PR #26, opened 2020-04-29
Simplify the Cincinnati Operator API

rthallisey:
Remove the CertConfigMapKey field and instead look for a hardcoded key,
'cincinnati-registry', in the openshift-config ConfigMap.  Having a
hardcoded value won't affect usability and it allows us to safely locate
the CA Cert.

The patch is a good segue into the future where we can potentially bundle
all the certs found in the openshift-config ConfigMap and mount them into
Cincinnati.

### DIFF
--- a/docs/external-registry-ca.md
+++ b/docs/external-registry-ca.md
@@ -10,8 +10,7 @@ that we'll use to store the external registry CA Cert.  You can read more about
 this API in the [OpenShift documentation](https://docs.openshift.com/container-platform/4.3/registry/configuring-registry-operator.html#images-configuration-cas_configuring-registry-operator).
 
 Create a ConfigMap in the `openshift-config` namespace.  Fill in your CA Cert
-and ConfigMap name, but keep the key, `cincinnati-registry`, because it's how
-Cincinnati locates your Cert:
+under the key `cincinnati-registry`, because it's how Cincinnati locates your Cert:
 ```yaml
 apiVersion: v1
 kind: ConfigMap
@@ -35,5 +34,5 @@ spec:
 ```
 
 The Cincinnati Operator will watch the `image.config.openshift.io` API and the
-`openshift-config` ConfigMap you created for changes and restart the deployment
-if the Cert has changed.
+ConfigMap you created in the `openshift-config` namespace for changes, then
+restart the deployment if the Cert has changed.
--- a/docs/external-registry-ca.md
+++ b/docs/external-registry-ca.md
@@ -1,0 +1,39 @@
+# External Registry CA Injection
+
+If you are using a secure external container registry to hold mirrored OpenShift
+release images, Cincinnati will need access to this registry in order to build
+an upgrade graph.  Here's how you can inject your CA Cert into the Cincinnati
+pod.
+
+OpenShift has an external registry API, located at `image.config.openshift.io`,
+that we'll use to store the external registry CA Cert.  You can read more about
+this API in the [OpenShift documentation](https://docs.openshift.com/container-platform/4.3/registry/configuring-registry-operator.html#images-configuration-cas_configuring-registry-operator).
+
+Create a ConfigMap in the `openshift-config` namespace.  Fill in your CA Cert
+and ConfigMap name, but keep the key, `cincinnati-registry`, because it's how
+Cincinnati locates your Cert:
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: trusted-ca
+data:
+  cincinnati-registry: |
+    -----BEGIN CERTIFICATE-----
+    ...
+    -----END CERTIFICATE-----
+```
+
+Edit the `cluster` resource from the `image.config.openshift.io` API and set
+the `additionalTrustedCA` field to the name of the ConfigMap you just created
+above.
+```bash
+$ oc edit image.config.openshift.io cluster
+spec:
+  additionalTrustedCA:
+    name: trusted-ca
+```
+
+The Cincinnati Operator will watch the `image.config.openshift.io` API and the
+`openshift-config` ConfigMap you created for changes and restart the deployment
+if the Cert has changed.
--- a/pkg/controller/cincinnati/cincinnati_controller.go
+++ b/pkg/controller/cincinnati/cincinnati_controller.go
@@ -217,22 +217,30 @@ func (r *ReconcileCincinnati) ensureAdditionalTrustedCA(ctx context.Context, req
 		return err
 	}
 
-	localCM := &corev1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      nameAdditionalTrustedCA(instance),
-			Namespace: instance.Namespace,
-		},
-		Data: sourceCM.Data,
-	}
+	if _, ok := sourceCM.Data[NameCertConfigMapKey]; ok {
+		localCM := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      nameAdditionalTrustedCA(instance),
+				Namespace: instance.Namespace,
+			},
+			Data: sourceCM.Data,
+		}
 
-	// Set Cincinnati instance as the owner and controller
-	if err := controllerutil.SetControllerReference(instance, localCM, r.scheme); err != nil {
-		return err
-	}
+		// Set Cincinnati instance as the owner and controller
+		if err := controllerutil.SetControllerReference(instance, localCM, r.scheme); err != nil {
+			return err
+		}
 
-	if err := r.ensureConfigMap(ctx, reqLogger, localCM); err != nil {
-		handleErr(reqLogger, &instance.Status, "EnsureConfigMapFailed", err)
-		return err
+		if err := r.ensureConfigMap(ctx, reqLogger, localCM); err != nil {
+			handleErr(reqLogger, &instance.Status, "EnsureConfigMapFailed", err)
+			return err
+		}
+		// Mount in ConfigMap data from the cincinnati-registry key
+		ExternalCACert := true
+		resources.graphBuilderContainer = resources.newGraphBuilderContainer(instance, r.operandImage, ExternalCACert)
+		resources.deployment = resources.newDeployment(instance, ExternalCACert)
+	} else {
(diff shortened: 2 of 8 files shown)


rthallisey:
Still need to add tests, but it's good for review.

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -231,22 +231,30 @@ func (r *ReconcileCincinnati) ensureAdditionalTrustedCA(ctx context.Context, req
...
 	}
 
-	localCM := &corev1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      nameAdditionalTrustedCA(instance),
-			Namespace: instance.Namespace,
-		},
-		Data: sourceCM.Data,
-	}
+	if _, ok := sourceCM.Data[NameCertConfigMapKey]; ok {
```
djzager, line 234:
```suggestion
	if _, ok := sourceCM.Data[NameCertConfigMapKey]; !ok {
		reqLogger.Info("Found ConfigMap referenced by ImageConfig.Spec.AdditionalTrustedCA.Name but did not find key 'cincinnati-registry' for registry CA cert.", "Name", image.Spec.AdditionalTrustedCA.Name, "Namespace", openshiftConfigNamespace)
		return nil
	}
	...		
```

djzager: Should we return an `error` in this scenario?

pkg/controller/cincinnati/names.go
```diff
@@ -14,6 +14,10 @@ const (
 	NameInitContainerGraphData string = "graph-data"
 	// openshiftConfigNamespace is the name of openshift's configuration namespace
 	openshiftConfigNamespace = "openshift-config"
+	// NameDeploymentTrustedCA is the name of the copied ConfigMap from openshift-config
+	NameDeploymentTrustedCA = "trusted-ca"
```
djzager, line 18: Calling this `NameDeployment...` is confusing given that it's always referring to the configmap.

```suggestion
	NameConfigMapTrustedCA = "trusted-ca"
```

docs/external-registry-ca.md
```diff
@@ -0,0 +1,39 @@
...
+release images, Cincinnati will need access to this registry in order to build
+an upgrade graph.  Here's how you can inject your CA Cert into the Cincinnati
+pod.
+
+OpenShift has an external registry API, located at `image.config.openshift.io`,
+that we'll use to store the external registry CA Cert.  You can read more about
+this API in the [OpenShift documentation](https://docs.openshift.com/container-platform/4.3/registry/configuring-registry-operator.html#images-configuration-cas_configuring-registry-operator).
+
+Create a ConfigMap in the `openshift-config` namespace.  Fill in your CA Cert
+and ConfigMap name, but keep the key, `cincinnati-registry`, because it's how
```
djzager, line 13: based on the implementation, it appears extremely important for the `ConfigMap` name to be `trusted-ca`.

pkg/controller/cincinnati/cincinnati_controller.go
rthallisey: My thought was not to return an error because this would be the case where a cluster has something in `ImageConfig.Spec.AdditionalTrustedCA.Name`, but the cincinnati-operator doesn't care about it.  I figured I would log a message in case a user did intend for Cincinnati to get the cert, but used the wrong key.

docs/external-registry-ca.md
rthallisey: You have to create a configmap with the same name that was added to `ImageConfig.Spec.AdditionalTrustedCA.Name`.  The operator watches for a configmap named whatever is in `ImageConfig.Spec.AdditionalTrustedCA.Name`, so it should accept other names.

pkg/controller/cincinnati/new.go
```diff
@@ -107,12 +107,13 @@ func newKubeResources(instance *cv1alpha1.Cincinnati, image string) (*kubeResour
 	if err != nil {
 		return nil, err
 	}
+	ExternalCACert := false
```
sacharya, line 110: Change the private variable to start with lowercase: `externalCACert`

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -231,22 +231,30 @@ func (r *ReconcileCincinnati) ensureAdditionalTrustedCA(ctx context.Context, req
...
-		handleErr(reqLogger, &instance.Status, "EnsureConfigMapFailed", err)
-		return err
+		if err := r.ensureConfigMap(ctx, reqLogger, localCM); err != nil {
+			handleErr(reqLogger, &instance.Status, "EnsureConfigMapFailed", err)
+			return err
+		}
+		// Mount in ConfigMap data from the cincinnati-registry key
+		ExternalCACert := true
+		resources.graphBuilderContainer = resources.newGraphBuilderContainer(instance, r.operandImage, ExternalCACert)
+		resources.deployment = resources.newDeployment(instance, ExternalCACert)
```
sacharya: Change the private variable to start with lowercase: `externalCACert`

Nit: WDYT about making the externalCACert a property of the ReconcileCincinnati struct, and initializing it to true in newReconciler() function or the beginning of the Reconcile? That way, you can create the newKubeResources() with externalCACert and not have to selectively override some of the resources.

rthallisey: I'm essentially copying around the same value, so I agree stylistically.   The second part of your comment: `not have to selectively override some of the resources`, would imply that once `ExternalCACert` changes to `true`, we need to trigger a Reconcile event.  That would make for a trade-off in that we won't be progressing forward because we would have to `return` after the change.  That might not be a big deal though, thoughts?

sacharya: Not sure I totally understand, but I'd check for r.externalCACert in ensureAdditionalCACert() and return immediately if it's set to false. I don't think we need the ConfigMap created if externalCACert is false.

docs/external-registry-ca.md
mhrivnak: I think this might read a little more clearly like this: "Fill in your CA Cert under the key `cincinnati-registry`." without talking about the name. I think it's implied below that the name of the confimap is up to you.

docs/external-registry-ca.md
```diff
@@ -0,0 +1,39 @@
...
+above.
+```bash
+$ oc edit image.config.openshift.io cluster
+spec:
+  additionalTrustedCA:
+    name: trusted-ca
+```
+
+The Cincinnati Operator will watch the `image.config.openshift.io` API and the
+`openshift-config` ConfigMap you created for changes and restart the deployment
```
mhrivnak, line 38: For clarity, I'd suggest "and the ConfigMap you created in the the `openshift-config` namespace for changes..."

pkg/controller/cincinnati/cincinnati_controller.go
mhrivnak: Maybe a good opportunity to use a condition in the status.

pkg/controller/cincinnati/names.go
```diff
@@ -14,6 +14,10 @@ const (
 	NameInitContainerGraphData string = "graph-data"
 	// openshiftConfigNamespace is the name of openshift's configuration namespace
 	openshiftConfigNamespace = "openshift-config"
+	// NameConfigMapTrustedCA is the name of the copied ConfigMap from openshift-config
```
mhrivnak, line 17: I think this should be a suffix appended to the `Cincinnati` instance's name, so we could theoretically continue to work with more than one Cincinnati deployed in the same namespace.

pkg/controller/cincinnati/cincinnati_controller.go
rthallisey: Right, I think we're saying similar things. Once `ensureAdditionalCACert()` notices that `r.externalCACert` has changed, it would have to wait until the next `Reconcile()` call to pick up the correct kubeResources and apply them.  I'm just curious if anyone thinks that's an anti-pattern as opposed to fixing and applying the kuberesources on the current `Reconcile()` call.

sacharya: Ok, ignore my comment. I didn't realize it was set to true only in the case when the configMap key was found.

pkg/controller/cincinnati/cincinnati_controller.go
mhrivnak: Condition could be a follow-up PR in case you want to get this merged and make an issue to track the condition.

rthallisey: https://github.com/openshift/cincinnati-operator/issues/29

pkg/controller/cincinnati/names.go
rthallisey: I think this goes along with https://github.com/openshift/cincinnati-operator/issues/27, which is about expanding how we discover the Cert.  Your suggestion is another path this could take, so I'll make a comment about it in the issue.  Would that suffice?

pkg/controller/cincinnati/cincinnati_controller.go
djzager: I think my comment about returning an error may have taken attention away from my actual comment. Instead of this huge `if { ... } else { log message }` just invert the okay and `return nil` if that's what you want to do.

rthallisey: Inverted it.  Let me know if I misunderstood the suggestion.

pkg/controller/cincinnati/names.go
mhrivnak: I think in either case, if we have two Cincinnati resources in the same namespace, we don't want them trying to "share" configmaps in any way. It's just hard in that case to keep track of who is the owner, should it get cleaned up or not when a particular Cincinnati resource is deleted, etc. It's always simpler to have each primary resource have its own independent set of secondary resources, and we can't do that if there's a name collision.

rthallisey: I thought you were talking about ` NameCertConfigMapKey` for some reason.  Ya, I'll change it to a suffix.

rthallisey: Looking this over again, `NameConfigMapTrustedCA` is only used as the name of the Volume mounts and not the ConfigMap - I'll change the name.  The source for these volume mounts is `nameAdditionalTrustedCA()` which uses `instance.Name + "-trusted-ca"` - I'll also change the hard coded string `"-trusted-ca"` to `NameCertConfigMapKey`.  In principle, I think what I have accomplishes the ask, but I'll make the aforementioned tweeks.

PR #28, opened 2020-04-29
Remove git fields from CRD

### LINKED ISSUE #25: remove git fields from CRD
The Cincinnati team wants to move away from having cincinnati obtain graph data directly from git, and instead distribute graph data as a container image. As such they prefer to remove git from the CRD and have the operator only focus on loading data via the init container.

For this task:
* remove fields related to git from the CRD
* remove those parts of the config template
* ensure that it's clear and easy how to make and utilize an init container

rwsu:
Fixes #25

### DIFF
--- a/deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
+++ b/deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
@@ -7,13 +7,4 @@ spec:
   registry: "quay.io"
   repository: "openshift-release-dev/ocp-release"
   certConfigMapKey: ""
-  # There two ways to specify the source of the Cincinnati graph data.
-  # The first is through a github repository specified by these three
-  # parameters.
-  gitHubOrg: "openshift"
-  gitHubRepo: "cincinnati-graph-data"
-  branch: "master"
-  # The second is through an init container specified by graphDataImage.
-  # When specified, graphDataImage takes precedence over gitHubOrg,
-  # gitHubRepo, and branch.
-  #graphDataImage: "your-registry/your-repo/your-init-container"
+  graphDataImage: "quay.io/rwsu/cincinnati-graph-data-container:latest"
--- a/deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
+++ b/deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
@@ -7,4 +7,4 @@ spec:
   registry: "quay.io"
   repository: "openshift-release-dev/ocp-release"
   certConfigMapKey: ""
-  graphDataImage: "quay.io/rwsu/cincinnati-graph-data-container:latest"
+  graphDataImage: "your-registry/your-repo/your-init-container"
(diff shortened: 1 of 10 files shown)


deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
```diff
@@ -7,13 +7,4 @@ spec:
...
-  # parameters.
-  gitHubOrg: "openshift"
-  gitHubRepo: "cincinnati-graph-data"
-  branch: "master"
-  # The second is through an init container specified by graphDataImage.
-  # When specified, graphDataImage takes precedence over gitHubOrg,
-  # gitHubRepo, and branch.
-  #graphDataImage: "your-registry/your-repo/your-init-container"
\ No newline at end of file
+  graphDataImage: "quay.io/rwsu/cincinnati-graph-data-container:latest"
```
rthallisey, line 10: Should we publish a default image in https://quay.io/repository/cincinnati ?

deploy/role_binding.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: cincinnati-operator
-  namespace: openshift-cincinnati
+  namespace: default
```
rthallisey, line 20: Looks like this was accidentally committed.  I think we want to keep 'openshift-cincinnati'.

deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
mhrivnak: I hesitate to create a situation where someone might start using an image we produce as a production mechanism for consuming OCP update data. I think we need to depend on OCP to deliver that critical data and not become redistributors of it.

So I think it's better to enable people to easily build their own. If we want to have our own build for test and dev purposes, we can perhaps name it something very obviously non-production quality like `unreliable-demo-graph-data` or something.

rthallisey: My comment is with respect to developers.  Denoting that in the name would be a good idea.

deploy/role_binding.yaml
rwsu: This was intentional. I wasn't able to deploy the operator as everything else in /deploy uses the default namespace. To use the openshift-cincinnati namespace I think the ServiceAccount and Deployment will also have to switch to openshift-cincinnati.

deploy/crds/cincinnati.openshift.io_v1alpha1_cincinnati_cr.yaml
rwsu: I'll revert the value back to "your-registry/your-repo/your-init-container". We can change it to something concrete once a non-production image is published somewhere.

deploy/role_binding.yaml
rthallisey: I'm not seeing a namespace designation on any of the objects.  Were you deploying in the default NS?

rwsu: I was using the default. I'll revert back to openshift-cincinnati. We'll need to update the docs so people know they need to deploy the operator to the openshift-cincinnati namespace.

PR #30, opened 2020-05-01
Create a route for Cincinnati service

sacharya:


sacharya:
Will add some tests, but wanted to leave it here for review.

pkg/controller/cincinnati/new.go
```diff
@@ -195,6 +198,32 @@ func (k *kubeResources) newPolicyEngineService(instance *cv1alpha1.Cincinnati) *
...
+				TargetPort: intstr.FromString("policy-engine"),
+			},
+			To: routev1.RouteTargetReference{
+				Kind: "Service",
+				Name: namePolicyEngineService(instance),
+			},
+			TLS: &routev1.TLSConfig{
+				Termination: routev1.TLSTerminationEdge,
+				InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyAllow,
+			},
```
sacharya, line 222: Any thoughts on how we should tls?

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -153,6 +154,7 @@ func (r *ReconcileCincinnati) Reconcile(request reconcile.Request) (reconcile.Re
 		r.ensureGraphBuilderService,
 		r.ensurePolicyEngineService,
 		r.ensurePodDisruptionBudget,
+		r.ensurePolicyEngineRoute,
```
sacharya, line 157: I only created route for the policy-engine, which exposes an HTTP API. It seems to me the graph-builder doesn't need to be exposed?

deploy/role.yaml
```diff
@@ -105,6 +105,17 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
rwsu, line 108: Does Role also need access to the routes resource?

pkg/controller/cincinnati/cincinnati_controller.go
rwsu: My understanding is graph-builder doesn't need to be exposed.

pkg/controller/cincinnati/new.go
rwsu: InsecureEdgeTerminationPolicyAllow is the default termination policy. The default is also used in the Cincinnati ingress route configuration doc: https://github.com/openshift/cincinnati/blob/master/docs/user/running-cincinnati.md

sacharya: With this configuration, we get a cert issued and signed by ingress-operator. So we are good there. Eg. 
```
Server certificate:
* 	subject: CN=*.apps.suachary-mycluster01.devcluster.openshift.com
* 	start date: May 05 18:49:04 2020 GMT
* 	expire date: May 05 18:49:05 2022 GMT
* 	common name: *.apps.suachary-mycluster01.devcluster.openshift.com
* 	issuer: CN=ingress-operator@1588704534
```

deploy/role.yaml
sacharya: We are using ClusterRole. There is another PR #37 for removing the Role resource altogether.

cmd/manager/main.go
```diff
@@ -129,6 +130,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := routev1.Install(mgr.GetScheme()); err != nil {
```
rwsu, line 133: I'm curious as to what this line is doing. I noticed when I commented out the equivalent in cincinnati_controller_test.go, a unit test failed with 

```
    TestEnsurePolicyEngineRoute/EnsurePolicyEngineRoute: cincinnati_controller_test.go:380: no kind is registered for the type v1.Route in scheme "pkg/runtime/scheme.go:101"
```

sacharya: That is the same as AddToScheme (which adds this version to the scheme), but AddToScheme is now deprecated.

pkg/controller/cincinnati/new.go
```diff
@@ -196,6 +199,32 @@ func (k *kubeResources) newPolicyEngineService(instance *cv1beta1.Cincinnati) *c
...
 }
 
+func (k *kubeResources) newPolicyEngineRoute(instance *cv1beta1.Cincinnati) *routev1.Route {
+	name := namePolicyEngineRoute(instance)
+	return &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: instance.Namespace,
+			Labels: map[string]string{
+				"app": name,
```
mhrivnak, line 209: For a `Cincinnati` with name `foo`, the route name would be `foo-policy-engine-route`, which is not exactly the "app" name. Perhaps this would be better if the value were the Deployment's name, which would then match the label on the Deployment.

sacharya: Done.

PR #38, opened 2020-05-11
Add some OLM testing tools

rthallisey:


tools/README.md
```diff
@@ -0,0 +1,41 @@
+# Tools
```
djzager, line 1: The `hack` directory is the conventional place for all things testing, scripts, etc. I think it should go there unless there is a convincing reason otherwise.

tools/operator-group.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: operators.coreos.com/v1
```
djzager, line 1: Manifests should go in `deploy` for operator-sdk based projects.

LalatenduMohanty:
@jottofar Do you have any plan for this PR?

jottofar:
@LalatenduMohanty Yes. But this PR is not required for the migration so would be after the migration "must have"s. But I have already created some tools that overlap with some of those contained within this PR as part of the migration effort.

PR #43, opened 2020-05-18
Functional testcases

pawanpinjarkar:
Functional test cases to create and verify deployment, replica sets, pod status, services, custom resource and PodDisruptionBudget.


functests/config.go
```diff
@@ -0,0 +1,271 @@
...
+		Kind:       "Cincinnati",
+	},
+	ObjectMeta: metav1.ObjectMeta{
+		Name: customResourceName,
+	},
+	Spec: cincinnativ1beta1.CincinnatiSpec{
+		Replicas:       1,
+		Registry:       "quay.io",
+		Repository:     "openshift-release-dev/ocp-release",
+		GraphDataImage: "quay.io/ppinjark/cincinnati-graph-data-container:latest",
```
pawanpinjarkar, line 269: The value needs to be changed

functests/config.go
```diff
@@ -0,0 +1,271 @@
...
+const (
+	customResourceName = "example-cincinnati"
+	operatorName       = "cincinnati-operator"
+	operatorNamespace  = "openshift-cincinnati"
+	defaultNamespace   = "default"
+	crdName            = "cincinnatis.cincinnati.openshift.io"
+	resource           = "cincinnatis"
+	replicas           = 1
+	retryInterval      = time.Second * 5
+	timeout            = time.Second * 1200
```
rwsu, line 23: I think 20 minutes is too long to wait for the deployment to come up. I think 5 minutes should be enough time.

functests/config.go
```diff
@@ -0,0 +1,271 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{
+					"name": operatorName,
+				},
+			},
+			Spec: v1.PodSpec{
+				ServiceAccountName: operatorName,
+				Containers: []v1.Container{{
+					Name:            operatorName,
+					Image:           "quay.io/cincinnati/cincinnati-operator:latest",
```
rwsu, line 190: This should point to the operator image built by the CI, not to quay.

functests/common.go
```diff
@@ -4,28 +4,252 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
-	k8sClient, err := kubernetes.NewForConfig(config)
+	return config, nil
+}
+
+// GetK8sClient is the function used to retrieve the kubernetes client
+func GetK8sClient() (*kubernetes.Clientset, error) {
+	config, err := GetConfig()
```
rwsu, line 37: When I don't have KUBECONFIG set in my environment. I see this error: 
````
$ make func-test
Running functional test suite
hack/functest.sh
GINKO binary found at /home/rwsu/go/bin/ginkgo
Compiling functests...
    compiled functests.test
panic: 
Your test failed.
Ginkgo panics to prevent subsequent assertions from running.
Normally Ginkgo rescues this panic so you shouldn't see it.

But, if you make an assertion in a goroutine, Ginkgo can't capture the panic.
To circumvent this, you should call

	defer GinkgoRecover()

at the top of the goroutine that caused this panic.


goroutine 1 [running]:
github.com/onsi/ginkgo.Fail(0xc0003ae580, 0x3d, 0xc00070eb40, 0x1, 0x1)
	/home/rwsu/go/pkg/mod/github.com/onsi/ginkgo@v1.12.0/ginkgo_dsl.go:267 +0xc8
github.com/onsi/gomega/internal/assertion.(*Assertion).match(0xc000729300, 0x170f5c0, 0x220a2b0, 0x0, 0x0, 0x0, 0x0, 0xc000729300)
	/home/rwsu/go/pkg/mod/github.com/onsi/gomega@v1.10.0/internal/assertion/assertion.go:79 +0x20e
github.com/onsi/gomega/internal/assertion.(*Assertion).NotTo(0xc000729300, 0x170f5c0, 0x220a2b0, 0x0, 0x0, 0x0, 0xc00009a060)
	/home/rwsu/go/pkg/mod/github.com/onsi/gomega@v1.10.0/internal/assertion/assertion.go:48 +0xc7
github.com/openshift/cincinnati-operator/functests.glob..func5()
	/home/rwsu/go/src/github.com/openshift/cincinnati-operator/functests/podDisruptionBudget_test.go:17 +0x131
github.com/onsi/ginkgo/internal/suite.(*Suite).PushContainerNode(0xc000146fa0, 0x152b9c6, 0x13, 0x15be4c0, 0x0, 0x20a51c7, 0x60, 0xd, 0xc00002c370, 0xa5)
	/home/rwsu/go/pkg/mod/github.com/onsi/ginkgo@v1.12.0/internal/suite/suite.go:144 +0x158
github.com/onsi/ginkgo.Describe(0x152b9c6, 0x13, 0x15be4c0, 0x0)
	/home/rwsu/go/pkg/mod/github.com/onsi/ginkgo@v1.12.0/ginkgo_dsl.go:294 +0xaa
github.com/openshift/cincinnati-operator/functests.init()
	/home/rwsu/go/src/github.com/openshift/cincinnati-operator/functests/podDisruptionBudget_test.go:13 +0x1d7

Ginkgo ran 1 suite in 3.214910501s
Test Suite Failed
make: *** [Makefile:12: func-test] Error 1
````

Comment (pawanpinjarkar) [reply functests/common.go:37]: Done

File: functests/config.go (hunk 2 above)
Comment (pawanpinjarkar) [reply functests/config.go]: Done

File: functests/config.go [hunk 5]
```diff
@@ -0,0 +1,271 @@
...
+	TypeMeta: metav1.TypeMeta{
+		APIVersion: "cincinnati.openshift.io/v1beta1",
+		Kind:       "Cincinnati",
+	},
+	ObjectMeta: metav1.ObjectMeta{
+		Name: customResourceName,
+	},
+	Spec: cincinnativ1beta1.CincinnatiSpec{
+		Replicas:       1,
+		Registry:       "quay.io",
```
pawanpinjarkar, line 267: @rwsu Please suggest if I need to change the registry to CI registry `registry.svc.ci.openshift.org`?

rwsu: No change

functests/config.go
pawanpinjarkar: Done

functests/config.go
pawanpinjarkar: Done

hack/functest.sh
```diff
@@ -2,12 +2,24 @@
 
 set -e
 
-# Override the image name when this is invoked from openshift ci
+OPERATOR_NAME="cincinnati-operator"
+GRAPH_DATA_CONTAINER_NAME="cincinnati-graph-data-container"
+REGISTRY="quay.io"
+OPERATOR_IMAGE="${REGISTRY}/cincinnati/${OPERATOR_NAME}:latest"
+GRAPH_DATA_IMAGE="${REGISTRY}/ppinjark/${GRAPH_DATA_CONTAINER_NAME}:latest"
```
djzager, line 9: Probably don't want to leave this pointing at your graph data image by default. Especially given that I don't believe it is overrideable in its current form.

pawanpinjarkar: The value gets replaced to `quay.io/ppinjark/cincinnati-graph-data-container:latest` when running locally. Not sure what other value we need to use. Is there a generic/standard value we should use here?

pawanpinjarkar: @djzager I have removed the dependency form the go config to create k8s resources. Now the resources are getting created from the deploy dir. Necessary replacements are made in the yaml files from the bash script hack/functest.sh

@sacharya @rwsu FYI

hack/functest.sh
```diff
@@ -2,10 +2,24 @@
...
+	sed -i "s|your-registry/your-repo/your-init-container|$GRAPH_DATA_IMAGE|" deploy/crds/cincinnati.openshift.io_v1beta1_cincinnati_cr.yaml
+
+else
+	if ! [ -n "$GRAPH_DATA_IMAGE" ]; then
+		echo "GRAPH_DATA_IMAGE environment variable must be set."
+		exit 1
+	fi
+	sed -i "s|quay.io/cincinnati/cincinnati:latest|quay.io/app-sre/cincinnati:2873c6b|" deploy/operator.yaml
+	sed -i "s|your-registry/your-repo/your-init-container|$GRAPH_DATA_IMAGE|" deploy/crds/cincinnati.openshift.io_v1beta1_cincinnati_cr.yaml
 fi
```
sacharya, line 23: Consider making two separate targets,`make deploy` and `make functest`.

functests/customResource_test.go
```diff
@@ -0,0 +1,125 @@
+package functests
```
sacharya, line 1: Rename the file to custom_resource_test.go

functests/deployment_test.go
```diff
@@ -0,0 +1,41 @@
...
+			Expect(err).To(BeNil())
+			Expect(string(pod.Items[0].Status.Phase)).Should(Equal("Running"))
+			Expect(bool(pod.Items[0].Status.ContainerStatuses[0].Ready)).Should(Equal(true))
+
+			By("Verifying if " + operatorName + "-metrics service is available")
+			err = WaitForService(k8sClient, operatorName+"-metrics")
+			Expect(err).To(BeNil())
+		})
+	})
+})
```
sacharya, line 41: I don't think there is much value in having this test. BeforeEach() step in the other test should cover this already.

sacharya: Sorry, not saying to dump everything in one test. Just that deployment_test is redundant, as custom_resource_test already needs deployment to be up and running. Running deployment_test after custom_resource_test doesn't make any sense, and these tests run in random order.

hack/functest.sh
sacharya: `make deploy` is useful outside of `make func-test`, when you just need the operator deployed.  And func-test would depend on deploy. Replacing values is an implementation detail on how the deploy is done.

functests/customResource_test.go
```diff
@@ -0,0 +1,125 @@
...
+var _ = Describe("CustomResource", func() {
+	RegisterFailHandler(Fail)
+
+	k8sClient, err := getK8sClient()
+	Expect(k8sClient).NotTo(BeNil())
+	Expect(err).To(BeNil())
+
+	BeforeEach(func() {
+		err = WaitForDeployment(k8sClient, operatorName)
+		Expect(err).To(BeNil())
```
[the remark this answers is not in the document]
pawanpinjarkar: I think it is a better idea to first "create" all the resources via `deploy` in the `BeforeSuite` and then let the test cases only focus on verifying the status of the objects it needs in the `BeforeEach`, when required. The purpose of these functional test cases as I envisioned, is to be able to confirm the objects the deployment creates so it is better to first "deploy everything" and then proceed on "testing". What is the benefit of your approach that I am not able to see?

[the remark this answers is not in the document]
sacharya: You only need to deploy the operator once. But you may want to create different CRs to test different things. And on every test, you want to create the resources it needs and clean it up at the end so the next test can start from a known state. Try writing a test to update replicas or graphDataImage.

functests/customResource_test.go
sacharya: Or better cincinnati_creation_test.go?

pawanpinjarkar: ok

hack/functest.sh
pawanpinjarkar: Ok, I can write a new target for deploy but then should creating a CR be not a part of the deploy? When we say to deploy the operator, I think it does mean creating a CR as well, please correct if I am wrong here. If I do not create the CR as a part of `make deploy` target then I think it is incomplete to say we have "deployed" the operator completely and then the purpose of that target is not fulfilled as then one will have to still manually create the CR. 

Regarding the (future) case if we have to create multiple CRs then the test case can be modified with BeforeEach and AfterEach at that time but as of now, I think I am covering all the functionalities we have in place currently. Please suggest if I am missing something that I am not testing for CR?

functests/customResource_test.go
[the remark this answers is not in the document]
pawanpinjarkar: I think this is not needed here ( a test to update replicas or graphDataImage etc) as that should be a part of the unit test and not functional tests.

functests/deployment_test.go
pawanpinjarkar: Agree that with the execution of test cases in random order, the situation when custom resource test runs prior to deployment test may look redundant but having a separate deployment test still helps in better readability and understanding which component should do what.

functests/customResource_test.go
pawanpinjarkar: Done

hack/functest.sh
pawanpinjarkar: Done.

functests/customResource_test.go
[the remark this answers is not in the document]
pawanpinjarkar: Done

functests/deployment_test.go
rwsu: I think it is also redundant.

pawanpinjarkar: Done

hack/deploy.sh
```diff
@@ -0,0 +1,35 @@
...
+set -e
+
+if [ -n "$OPENSHIFT_BUILD_NAMESPACE" ]; then
+	OPERATOR_IMAGE="registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:cincinnati-operator"
+	GRAPH_DATA_IMAGE="registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:cincinnati-graph-data-container"
+
+	echo "Openshift CI detected, deploying using image $OPERATOR_IMAGE and ${GRAPH_DATA_IMAGE}"
+
+	sed -i "s|quay.io/cincinnati/cincinnati-operator:latest|$OPERATOR_IMAGE|" deploy/operator.yaml
+	sed -i "s|quay.io/cincinnati/cincinnati:latest|quay.io/app-sre/cincinnati:2873c6b|" deploy/operator.yaml
```
sacharya, line 12: Make this an env variable defaulting to `quay.io/app-sre/cincinnati:2873c6b`?

hack/deploy.sh
```diff
@@ -0,0 +1,35 @@
...
+	if ! [ -n "$KUBECONFIG" ]; then
+		echo "KUBECONFIG environment variable must be set."
+		exit 1
+	fi
+	if ! [ -n "$GRAPH_DATA_IMAGE" ]; then
+		echo "GRAPH_DATA_IMAGE environment variable must be set."
+		exit 1
+	fi
+	sed -i "s|quay.io/cincinnati/cincinnati:latest|quay.io/app-sre/cincinnati:2873c6b|" deploy/operator.yaml
+	sed -i "s|your-registry/your-repo/your-init-container|$GRAPH_DATA_IMAGE|" deploy/crds/cincinnati.openshift.io_v1beta1_cincinnati_cr.yaml
```
sacharya, line 25: Also consider making OPERATOR_IMAGE a variable that can be overridden. And since the sed statements are the same for both if and else,you could move those outside the if-else.

hack/deploy.sh
pawanpinjarkar: Done but with another way which is more in line with the ReadMe and code.

hack/deploy.sh
pawanpinjarkar: Done

hack/deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+
+DEFAULT_OPERATOR_IMAGE="quay.io/cincinnati/cincinnati-operator:latest"
+DEFAULT_OPERAND_IMAGE="quay.io/cincinnati/cincinnati:latest"
+OPERAND_IMAGE="quay.io/app-sre/cincinnati:2873c6b"
+
+if ! [ -n "$CUSTOM_OPERAND_IMAGE" ]; then
+	sed -i "s|$DEFAULT_OPERAND_IMAGE|$OPERAND_IMAGE|" deploy/operator.yaml
+else
+	sed -i "s|$DEFAULT_OPERAND_IMAGE|$CUSTOM_OPERAND_IMAGE|" deploy/operator.yaml
+fi
```
sacharya, line 13: You could do something like:
```OPERAND_IMAGE="{$OPERAND_IMAGE:-${DEFAULT_OPERAND_IMAGE}}"```

and move the sed statements outside of each if-else, as they are applicable in each case.
```
sed -i "s|$DEFAULT_OPERAND_IMAGE|$OPERAND_IMAGE|" deploy/operator.yaml
sed -i "s|$DEFAULT_OPERATOR_IMAGE|$OPERATOR_IMAGE|" deploy/operator.yaml
sed -i "s|your-registry/your-repo/your-init-container|$GRAPH_DATA_IMAGE|" deploy/crds/cincinnati.openshift.io_v1beta1_cincinnati_cr.yaml
```

PR #46, opened 2020-05-27
Allow user to change the cert on the route outside of the CR

sacharya:
The route is configured with a default cert from the ingress. If the customer wants to use a different TLS cert/key pair,this patch allows them to manually modify the route object with their cert/key and the controller won't override the change. If the CR or the route is deleted, the customer would have to remember to make that change again.

Alternatively, we could change our API to ask for certificate, private key and CA certificate. Other option is to ask the customer to create a new route themselves with the certificate they want.

### DIFF
--- a/pkg/controller/cincinnati/cincinnati_controller.go
+++ b/pkg/controller/cincinnati/cincinnati_controller.go
@@ -494,10 +494,13 @@ func (r *ReconcileCincinnati) ensurePolicyEngineRoute(ctx context.Context, reqLo
 	}
 
 	// found existing resource; let's compare and update if needed
-	if !reflect.DeepEqual(found.Spec, route.Spec) {
+	if !reflect.DeepEqual(found.Spec.Port, route.Spec.Port) || !reflect.DeepEqual(found.Spec.To, route.Spec.To) {
 		reqLogger.Info("Updating Route", "Namespace", route.Namespace, "Name", route.Name)
 		updated := found.DeepCopy()
-		updated.Spec = route.Spec
+		// Update everything but updated.Spec.TLS. We want to allow user to update the TLS cert/key manually on the route
+		// and we don't want to override that change.
+		updated.Spec.Port = route.Spec.Port
+		updated.Spec.To = route.Spec.To
 		err = r.client.Update(ctx, updated)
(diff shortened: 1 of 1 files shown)


pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -494,10 +494,13 @@ func (r *ReconcileCincinnati) ensurePolicyEngineRoute(ctx context.Context, reqLo
 	}
 
 	// found existing resource; let's compare and update if needed
-	if !reflect.DeepEqual(found.Spec, route.Spec) {
+	if !reflect.DeepEqual(found.Spec.Port, route.Spec.Port) || !reflect.DeepEqual(found.Spec.To, route.Spec.To) {
 		reqLogger.Info("Updating Route", "Namespace", route.Namespace, "Name", route.Name)
 		updated := found.DeepCopy()
-		updated.Spec = route.Spec
+		// Update everything but updated.Spec.TLS. We want to allow user to update the TLS cert/key manually on the route
```
mhrivnak, line 500: Just sanity-checking; `TLS` is not the only field being ignored, right? Maybe it's worth a quick doc statement on why these are the only two fields we want to manage after creation?

sacharya: Port, To and TLS are the only things we configure. See. https://github.com/openshift/cincinnati-operator/blob/master/pkg/controller/cincinnati/new.go#L212-L222

Here's the output from the Route object. It adds the default `host` field and `wildcardPolicy` of None.
```
spec:
  host: example-cincinnati-policy-engine-route-openshift-cincinnati.apps-crc.testing
  port:
    targetPort: policy-engine
  tls:
    insecureEdgeTerminationPolicy: Allow
    termination: edge
  to:
    kind: Service
    name: example-cincinnati-policy-engine
    weight: 100
  wildcardPolicy: None
```

mhrivnak: Gotcha. So here by "everything" you mean "everything this operator originally set", not "everything in the Spec". I think it would be marginally more clear to make that distinction explicitly, but it's up to you.

PR #52, opened 2020-07-07
Added support for choosing bind address

mvazquezc:
As of today, the operator will always use "0.0.0.0" as the binding address for the different services. When running on IPv6 environments that cause the services to fail since there are not IPv4 addresses available in the pod. With this change, the user will be able to choose which address is used when binding. 

By default, if no value is provided for the `address` property in the spec, we will configure "0.0.0.0", otherwise, we will configure the address configured by the user. 

Let me know what you think about the implementation, if it looks good to you I'll update the docs accordingly.

### DIFF
--- a/pkg/apis/cincinnati/v1beta1/cincinnati_types.go
+++ b/pkg/apis/cincinnati/v1beta1/cincinnati_types.go
@@ -25,13 +25,6 @@ type CincinnatiSpec struct {
 	// GraphDataImage is a container image that contains the Cincinnati graph
 	// data. The data is copied to /var/lib/cincinnati/graph-data.
 	GraphDataImage string `json:"graphDataImage"`
-
-	// IPv6 defines whether the services bind to "0.0.0.0" or to "::"
-	IPv6 bool `json:"ipv6,omitempty"`
-
-	// Address internally used for configuring the ConfigMap templates based on
-	// the IPv6 property
-	Address string
 }
 
 // CincinnatiStatus defines the observed state of Cincinnati
--- a/pkg/apis/cincinnati/v1beta1/cincinnati_types.go
+++ b/pkg/apis/cincinnati/v1beta1/cincinnati_types.go
@@ -25,6 +25,9 @@ type CincinnatiSpec struct {
 	// GraphDataImage is a container image that contains the Cincinnati graph
 	// data. The data is copied to /var/lib/cincinnati/graph-data.
 	GraphDataImage string `json:"graphDataImage"`
+
+	// Address defines which address should be used when services bind to the network
+	Address string `json:"address,omitempty"`
 }
 
 // CincinnatiStatus defines the observed state of Cincinnati
--- a/pkg/apis/cincinnati/v1beta1/cincinnati_types.go
+++ b/pkg/apis/cincinnati/v1beta1/cincinnati_types.go
@@ -26,8 +26,12 @@ type CincinnatiSpec struct {
 	// data. The data is copied to /var/lib/cincinnati/graph-data.
 	GraphDataImage string `json:"graphDataImage"`
 
-	// Address defines which address should be used when services bind to the network
-	Address string `json:"address,omitempty"`
+	// IPv6 defines whether the services bind to "0.0.0.0" or to "::"
+	IPv6 bool `json:"ipv6,omitempty"`
+
+	// Address internally used for configuring the ConfigMap templates based on
+	// the IPv6 property
+	Address string
 }
 
 // CincinnatiStatus defines the observed state of Cincinnati
--- a/pkg/controller/cincinnati/new.go
+++ b/pkg/controller/cincinnati/new.go
@@ -36,11 +36,11 @@ const graphBuilderTOML string = `verbosity = "vvv"
 
 [service]
 pause_secs = 300
-address = "0.0.0.0"
+address = "{{.Address}}"
 port = 8080
 
 [status]
-address = "0.0.0.0"
+address = "{{.Address}}"
 port = 9080
 
 [[plugin_settings]]
@@ -225,6 +225,11 @@ func (k *kubeResources) newPolicyEngineRoute(instance *cv1beta1.Cincinnati) *rou
 }
 
 func (k *kubeResources) newEnvConfig(instance *cv1beta1.Cincinnati) *corev1.ConfigMap {
+	// If no value is provided in the spec for the bind address
+	// we default to all IPv4 addresses
+	if instance.Spec.Address == "" {
+		instance.Spec.Address = "0.0.0.0"
+	}
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      nameEnvConfig(instance),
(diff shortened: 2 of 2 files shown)


pkg/apis/cincinnati/v1beta1/cincinnati_types.go
```diff
@@ -25,6 +25,9 @@ type CincinnatiSpec struct {
 	// GraphDataImage is a container image that contains the Cincinnati graph
 	// data. The data is copied to /var/lib/cincinnati/graph-data.
 	GraphDataImage string `json:"graphDataImage"`
+
+	// Address defines which address should be used when services bind to the network
```
rthallisey, line 29: We could do a bool for ipv6 and based on it's value, set the address to `0.0.0.0` or `::/0`. WDYT?

pkg/apis/cincinnati/v1beta1/cincinnati_types.go
```diff
@@ -25,6 +25,9 @@ type CincinnatiSpec struct {
...
+	Address string `json:"address,omitempty"`
```
rthallisey, line 30: The operator metadata in `deploy` will also need to be updated.

pkg/apis/cincinnati/v1beta1/cincinnati_types.go
mvazquezc: That was my first approach. But then since you're using a template for building the configmap I didn't know how to translate that to the template. If you provide the guidance I'll go that way.

mvazquezc: Added IPv6 boolean to the Spec.

pkg/apis/cincinnati/v1beta1/cincinnati_types.go
mvazquezc: Will do

pkg/apis/cincinnati/v1beta1/cincinnati_types.go
```diff
@@ -25,6 +25,13 @@ type CincinnatiSpec struct {
 	// GraphDataImage is a container image that contains the Cincinnati graph
 	// data. The data is copied to /var/lib/cincinnati/graph-data.
 	GraphDataImage string `json:"graphDataImage"`
+
+	// IPv6 defines whether the services bind to "0.0.0.0" or to "::"
+	IPv6 bool `json:"ipv6,omitempty"`
```
djzager, line 30: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

My gut feeling is that adding the `Address` to the spec is sufficient. That way we can default to '0.0.0.0' or force it to be defined.

mhrivnak:
Is there ever a circumstance where we wouldn't want it to bind to both if available? Seems like that would be a sensible default. Is there a way to make cincinnati itself do that?

pkg/apis/cincinnati/v1beta1/cincinnati_types.go
mvazquezc: The first implementation was done in that way, then I changed it based on Ryan's suggestions, anyway I'd say a boolean makes sense, a user won't likely know which address is going to get the pod when created, and between defining a property with the bind-address or just setting a boolean to decide if the services listen on IPv6 or IPv4 I'd rather go with the second option, but again, I'm open to suggestions.

pkg/controller/cincinnati/new.go
```diff
@@ -225,18 +225,24 @@ func (k *kubeResources) newPolicyEngineRoute(instance *cv1beta1.Cincinnati) *rou
 }
 
 func (k *kubeResources) newEnvConfig(instance *cv1beta1.Cincinnati) *corev1.ConfigMap {
+	// If IPv6 boolean property is not configured in the Spec
```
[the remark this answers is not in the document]
mvazquezc: I added Address to the spec because the spec is used for rendering the template for one of the ConfigMaps, I didn't know if there is a better way to do that, but in this case, we only use that field for rendering the CM, I don't see any value on persisting its value in the api-server. Maybe there is a better solution?

mvazquezc:
> Is there ever a circumstance where we wouldn't want it to bind to both if available? Seems like that would be a sensible default. Is there a way to make cincinnati itself do that?

When you say both, do you mean IPv4 and IPv6? 

Dual stack is not supported yet, but in case you could have a pod with an IPv4 and an IPv6 address binding to one of the two will be enough for Cincinnati to work based on what I've seen. 

In this case the tricky thing will be knowing if the cluster has dual-stack support.

pkg/apis/cincinnati/v1beta1/cincinnati_types.go
djzager: I don't understand.

> a user won't likely know which address is going to get the pod when created

Then why would we bother taking the `Address` on the spec?

mhrivnak:
Yes, I mean both ipv4 and ipv6. As an example, both nginx and httpd allow you to either specify just a port or `*:<port>` to listen on, and then they will listen on all addresses whether ipv4 or ipv6. Can we just enable Cincinnati to do that?

It seems a shame to add options to the operator's API if there's no user story other than "As a user, I want cincinnati to listen on all addresses." If we need to do something short-term in the operator to make this work on ipv6 while waiting for cincinnati to support dual-stack, maybe we can come up with something that doesn't involve adding to the API.

rthallisey:
> Yes, I mean both ipv4 and ipv6. As an example, both nginx and httpd allow you to either specify just a port or *:<port> to listen on, and then they will listen on all addresses whether ipv4 or ipv6. Can we just enable Cincinnati to do that?
>
I agree.  Cincy is the only thing that's listening in this address space, so it would just make sense to support everything without exposing it.

I think we would need to change the default setting in both the operator and in Cincinnati. Setting the default to `::` here should cover it on this end.  I'll file a ticket on cincy as well.

pkg/apis/cincinnati/v1beta1/cincinnati_types.go
mvazquezc: I mean, if you add the Address to the spec so users can choose which address the services bind to, they will likely end using "0.0.0.0" or "::", the IP address that the SDN will assign to the pod is not something you can know beforehand easily. That's why I thought a boolean makes more sense.

mvazquezc:
Hey, as Ryan suggested. Binding to "::" works for IPv6 and IPv4 environments. I just reverted the changes on the API Types and edited the configmaps created by the controller.

pkg/apis/cincinnati/v1beta1/cincinnati_types.go
mvazquezc: Binding to "::" works on IPv4 and IPv6 clusters. I reverted the changes from the API types.

pkg/controller/cincinnati/new.go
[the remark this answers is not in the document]
mvazquezc: Binding to "::" works on IPv4 and IPv6 clusters. I reverted the changes from the API types.

mhrivnak:
I'm surprised that `::` includes ipv4 addresses, since it is ipv6 notation. But if it works, it works. It would be a good idea to test on a single-stack ipv4 system to be sure, but generally I think this is a lot better than adding API options.

rthallisey:
I was able to see it work on my local ipv4 env.  

This was where I got the idea from originally: https://github.com/nodejs/node/issues/9390.  You can disable the duel-stack listening by configuring `net.ipv6.bindv6only`.

Without knowing more about what admins do with `net.ipv6.bindv6only`, it's hard for me to say if we'll run into an issue or not.  I think for now, we can say this works as is, but I'll link this PR in the OTA duel stack jira card so it's on their radar when they do a more in depth analysis of this feature for 4.6.

PR #53, opened 2020-07-15
Copy pull-secret into graphbuilder

rthallisey:
The [basic auth support](https://github.com/openshift/cincinnati/pull/292) patch in Cincinnati added a `credentials_path` to point to a docker config file with basic auth for registries.  Since it's likely the installer is using the same local registry as CIncinnati, we'll mount the `pull-secret` secret into the graph builder.  This avoids the user having to essentially create the pull-secret twice.

### DIFF
--- a/pkg/controller/cincinnati/cincinnati_controller.go
+++ b/pkg/controller/cincinnati/cincinnati_controller.go
@@ -235,15 +235,15 @@ func handleCACertStatus(reqLogger logr.Logger, status *cv1beta1.CincinnatiStatus
 }
 
 func (r *ReconcileCincinnati) postAddPullSecret(ctx context.Context, reqLogger logr.Logger, instance *cv1beta1.Cincinnati, resources *kubeResources) error {
-	sourcePS, err := r.findPullSecret(ctx, reqLogger, instance, resources)
-	if err != nil {
+	sourcePS := &corev1.Secret{}
+	err := r.client.Get(ctx, types.NamespacedName{Name: namePullSecret, Namespace: openshiftConfigNamespace}, sourcePS)
+	if err != nil && errors.IsNotFound(err) {
+		handleErr(reqLogger, &instance.Status, "PullSecretNotFound", err)
 		return err
-	} else if sourcePS == nil {
-		return nil
+	} else if err != nil {
+		return err
 	}
-
-	resources.newPullSecret(instance, sourcePS)
-	resources.addPullSecret(instance)
+	resources.addPullSecret(instance, sourcePS)
 
 	return nil
 }
@@ -257,25 +257,9 @@ func (r *ReconcileCincinnati) postAddExternalCACert(ctx context.Context, reqLogg
 		return nil
 	}
 
-	resources.newTrustedCAConfig(instance, sourceCM)
-	resources.addExternalCACert(instance)
+	resources.addExternalCACert(instance, sourceCM)
 
 	return nil
-}
-
-// findPullSecret - Locate the pull secret in openshift-config and return it
-func (r *ReconcileCincinnati) findPullSecret(ctx context.Context, reqLogger logr.Logger, instance *cv1beta1.Cincinnati, resources *kubeResources) (*corev1.Secret, error) {
-	// Search for the the pull-secret in openshift-config
-	pullSecret := &corev1.Secret{}
-	err := r.client.Get(ctx, types.NamespacedName{Name: NamePullSecret, Namespace: openshiftConfigNamespace}, pullSecret)
-	if err != nil && errors.IsNotFound(err) {
-		handleErr(reqLogger, &instance.Status, "PullSecretNotFound", err)
-		return nil, err
-	} else if err != nil {
-		return nil, err
-	}
-
-	return pullSecret, nil
 }
 
 // findTrustedCAConfig - Locate the ConfigMap referenced by the ImageConfig resource in openshift-config and return it
@@ -320,13 +304,6 @@ func (r *ReconcileCincinnati) findTrustedCAConfig(ctx context.Context, reqLogger
 }
 
 func (r *ReconcileCincinnati) ensurePullSecret(ctx context.Context, reqLogger logr.Logger, instance *cv1beta1.Cincinnati, resources *kubeResources) error {
-	sourcePS, err := r.findPullSecret(ctx, reqLogger, instance, resources)
-	if err != nil {
-		return err
-	} else if sourcePS == nil {
-		return nil
-	}
-
 	// Set Cincinnati instance as the owner and controller
 	if err := controllerutil.SetControllerReference(instance, resources.pullSecret, r.scheme); err != nil {
 		return err
--- a/pkg/controller/cincinnati/cincinnati_controller.go
+++ b/pkg/controller/cincinnati/cincinnati_controller.go
@@ -24,8 +24,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	apicfgv1 "github.com/openshift/api/config/v1"
-	cv1beta1 "github.com/openshift/cincinnati-operator/pkg/apis/cincinnati/v1beta1"
 	routev1 "github.com/openshift/api/route/v1"
+	cv1beta1 "github.com/openshift/cincinnati-operator/pkg/apis/cincinnati/v1beta1"
 	"github.com/openshift/cluster-image-registry-operator/pkg/defaults"
 )
 
@@ -120,10 +120,10 @@ type ReconcileCincinnati struct {
 // and what is in the Cincinnati.Spec
 func (r *ReconcileCincinnati) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	/*    **Reconcile Pattern**
-	  1. Cache all the kubeResources
-	  2. Make a few modifications to the kubeResources
-	  3. Create all kubeResource
-	  4. Ensure all the kubeResources are correct
+	1. Create all the kubeResources
+	2. Make a few modifications to the kubeResources
+	3. Regenerate some of the kubeResources
+	4. Ensure all the kubeResources are correct in the Cluster
 	*/
 
 	ctx := context.TODO()
@@ -161,9 +161,11 @@ func (r *ReconcileCincinnati) Reconcile(request reconcile.Request) (reconcile.Re
 	} {
 		err = f(ctx, reqLogger, instanceCopy, resources)
 		if err != nil {
-			break
+			return reconcile.Result{}, err
 		}
 	}
+	// Regenerate resources to pickup any supplemental changes
+	resources.regenerate(instance)
 
 	conditionsv1.SetStatusCondition(&instanceCopy.Status.Conditions, conditionsv1.Condition{
 		Type:    cv1beta1.ConditionReconcileCompleted,
@@ -235,9 +237,9 @@ func handleCACertStatus(reqLogger logr.Logger, status *cv1beta1.CincinnatiStatus
 func (r *ReconcileCincinnati) postAddPullSecret(ctx context.Context, reqLogger logr.Logger, instance *cv1beta1.Cincinnati, resources *kubeResources) error {
 	sourcePS, err := r.findPullSecret(ctx, reqLogger, instance, resources)
 	if err != nil {
-	   return err
+		return err
 	} else if sourcePS == nil {
-	   return nil
+		return nil
 	}
 
 	resources.newPullSecret(instance, sourcePS)
@@ -250,9 +252,9 @@ func (r *ReconcileCincinnati) postAddExternalCACert(ctx context.Context, reqLogg
 	// Search for the the pull-secret in openshift-config
 	sourceCM, err := r.findTrustedCAConfig(ctx, reqLogger, instance, resources)
 	if err != nil {
-	   return err
+		return err
 	} else if sourceCM == nil {
-	   return nil
+		return nil
 	}
 
 	resources.newTrustedCAConfig(instance, sourceCM)
@@ -268,10 +270,11 @@ func (r *ReconcileCincinnati) findPullSecret(ctx context.Context, reqLogger logr
 	err := r.client.Get(ctx, types.NamespacedName{Name: NamePullSecret, Namespace: openshiftConfigNamespace}, pullSecret)
 	if err != nil && errors.IsNotFound(err) {
 		handleErr(reqLogger, &instance.Status, "PullSecretNotFound", err)
-		return nil, nil
+		return nil, err
 	} else if err != nil {
 		return nil, err
 	}
+
 	return pullSecret, nil
 }
 
@@ -319,9 +322,9 @@ func (r *ReconcileCincinnati) findTrustedCAConfig(ctx context.Context, reqLogger
 func (r *ReconcileCincinnati) ensurePullSecret(ctx context.Context, reqLogger logr.Logger, instance *cv1beta1.Cincinnati, resources *kubeResources) error {
 	sourcePS, err := r.findPullSecret(ctx, reqLogger, instance, resources)
 	if err != nil {
-	   return err
+		return err
 	} else if sourcePS == nil {
-	   return nil
+		return nil
 	}
 
 	// Set Cincinnati instance as the owner and controller
@@ -340,9 +343,9 @@ func (r *ReconcileCincinnati) ensurePullSecret(ctx context.Context, reqLogger lo
 func (r *ReconcileCincinnati) ensureAdditionalTrustedCA(ctx context.Context, reqLogger logr.Logger, instance *cv1beta1.Cincinnati, resources *kubeResources) error {
 	sourceCM, err := r.findTrustedCAConfig(ctx, reqLogger, instance, resources)
 	if err != nil {
-	   return err
+		return err
 	} else if sourceCM == nil {
-	   return nil
+		return nil
 	}
 
 	conditionsv1.SetStatusCondition(&instance.Status.Conditions, conditionsv1.Condition{
@@ -361,10 +364,6 @@ func (r *ReconcileCincinnati) ensureAdditionalTrustedCA(ctx context.Context, req
 		handleErr(reqLogger, &instance.Status, "EnsureConfigMapFailed", err)
 		return err
 	}
-
-	// Mount in ConfigMap data from the cincinnati-registry key
-	resources.addExternalCACert(instance)
-
 	return nil
 }
 
@@ -373,10 +372,10 @@ func (r *ReconcileCincinnati) ensureDeployment(ctx context.Context, reqLogger lo
 	if err := controllerutil.SetControllerReference(instance, deployment, r.scheme); err != nil {
 		return err
 	}
-
 	// Check if this deployment already exists
 	found := &appsv1.Deployment{}
 	err := r.client.Get(ctx, types.NamespacedName{Name: deployment.Name, Namespace: deployment.Namespace}, found)
+
 	if err != nil && errors.IsNotFound(err) {
 		reqLogger.Info("Creating Deployment", "Namespace", deployment.Namespace, "Name", deployment.Name)
 		err := r.client.Create(ctx, deployment)
--- a/pkg/controller/cincinnati/cincinnati_controller.go
+++ b/pkg/controller/cincinnati/cincinnati_controller.go
@@ -146,15 +146,25 @@ func (r *ReconcileCincinnati) Reconcile(request reconcile.Request) (reconcile.Re
 	instanceCopy := instance.DeepCopy()
 	instanceCopy.Status = cv1beta1.CincinnatiStatus{}
 
-	// this object creates all the kube resources we need and then holds them as
-	// the canonical reference for those resources during reconciliation.
+	// Start construction of resources
+
+	// 1. Create all the kubeResources
+	//    'newKubeResources' creates all the kube resources we need and holds
+	//    them in 'resources' as the canonical reference for those resources
+	//    during reconciliation.
 	resources, err := newKubeResources(instanceCopy, r.operandImage)
 	if err != nil {
 		reqLogger.Error(err, "Failed to render resources")
 		return reconcile.Result{}, err
 	}
 
-	// Supplemental kube resources changes
+	// 2. Make a few modifications to the kubeResources
+	//    The supplemental kube resources changes are modifications that are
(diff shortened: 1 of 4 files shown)


mhrivnak:
I'm not following why a local registry cares about the pull secret. Can you elaborate on the use case?

rthallisey:
fixed the description

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -560,3 +597,25 @@ func (r *ReconcileCincinnati) ensureConfigMap(ctx context.Context, reqLogger log
 
 	return nil
 }
+
+func (r *ReconcileCincinnati) ensureSecret(ctx context.Context, reqLogger logr.Logger, cm *corev1.Secret) error {
```
mhrivnak, line 601: `cm` was an abbreviation of `ConfigMap` in the `ensureConfigMap` function, but perhaps here we should use something else. `secret` may be sufficient.

pkg/controller/cincinnati/new.go
```diff
@@ -333,6 +334,31 @@ func (k *kubeResources) newDeployment(instance *cv1beta1.Cincinnati) *appsv1.Dep
...
+
+	// Add the volumeMount to the graph builder container
+	k.graphBuilderContainer.VolumeMounts = append(k.graphBuilderContainer.VolumeMounts, corev1.VolumeMount{
+			Name:      NamePullSecret,
+			ReadOnly:  true,
+			MountPath: "/var/lib/cincinnati/registry-credentials",
+	})
+
+	// Rebuild the deployment since we modified the container definition
+	k.deployment = k.newDeployment(instance)
```
mhrivnak, line 348: Rather than re-create this, it would be better to find a way to know up-front, before `newKubeResources()` gets called, whether this volume mount should be added. It looks like you followed the pattern from ensuring the configmap exists, and I think the same applies there. I'm thinking the `Reconcile` function should:

1. ensure the configmap exists in the namespace as appropriate
2. ensure the secret exists in the namespace as appropriate
3. call `newKubeResources` and let it add volume mounts for the configmap and secret if they exist
4. carry on with the rest of the "ensure" functions

That way we don't have to worry about re-creating resources, tracking whether anything else has already done anything with it in the meantime, etc.

WDYT? cc @djzager

rthallisey:
Some background on why I'm targeting the pull secret: https://docs.openshift.com/container-platform/4.2/installing/install_config/installing-restricted-networks-preparations.html#installation-adding-registry-pull-secret_installing-restricted-networks-preparations

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -214,6 +215,42 @@ func handleCACertStatus(reqLogger logr.Logger, status *cv1beta1.CincinnatiStatus
 	reqLogger.Info(message)
 }
 
+func (r *ReconcileCincinnati) ensurePullSecret(ctx context.Context, reqLogger logr.Logger, instance *cv1beta1.Cincinnati, resources *kubeResources) error {
+
+	// Search for the the pull-secret in openshift-config
+	pullSecret := &corev1.Secret{}
+	err := r.client.Get(ctx, types.NamespacedName{Name: NamePullSecret, Namespace: openshiftConfigNamespace}, pullSecret)
+	if err != nil && errors.IsNotFound(err) {
+		handleErr(reqLogger, &instance.Status, "PullSecretNotFound", err)
```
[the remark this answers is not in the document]
rthallisey: Afaik, you'll always have a pull-secret Secret in your cluster, so I think it's safe to always error.

Ran a few tests to verify:
```
$ openshift-install create cluster
FATAL failed to fetch Terraform Variables: failed to load asset "Install Config": invalid "install-config.yaml" file: pullSecret: 
Invalid value: "": unexpected end of JSON input 
```
```
$ openshift-install create cluster
FATAL failed to fetch Terraform Variables: failed to load asset "Install Config": invalid "install-config.yaml" file: pullSecret: Invalid value: "{}": auths required
 ```

And from the code:
```go
        if len(s.Auths) == 0 {
                return fmt.Errorf("auths required")
        }
        errs := []error{}
        for d, a := range s.Auths {
                _, authPresent := a["auth"]
                _, credsStorePresnet := a["credsStore"]
                if !authPresent && !credsStorePresnet {
                        errs = append(errs, fmt.Errorf("%q requires either auth or credsStore", d))
                }
        }
        return k8serrors.NewAggregate(errs)
```

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -214,40 +232,117 @@ func handleCACertStatus(reqLogger logr.Logger, status *cv1beta1.CincinnatiStatus
...
+func (r *ReconcileCincinnati) postAddPullSecret(ctx context.Context, reqLogger logr.Logger, instance *cv1beta1.Cincinnati, resources *kubeResources) error {
+	sourcePS, err := r.findPullSecret(ctx, reqLogger, instance, resources)
+	if err != nil {
+	   return err
+	} else if sourcePS == nil {
+	   return nil
+	}
+
+	resources.newPullSecret(instance, sourcePS)
+	resources.addPullSecret(instance)
```
mhrivnak, line 244: I'm hoping we can avoid modifying `resources` after instantiating it. What I'm proposing is that `ensurePullSecret` runs before `newKubeResources`. That way, when all the resources are being generated, it will already be able to consume the pull secret copy. WDYT? Does that make sense?

We would apply the same pattern to `ensureAdditionalTrustedCA`, even though I know that's not part of this PR.

rthallisey: > I'm hoping we can avoid modifying resources after instantiating it. What I'm proposing is that ensurePullSecret runs before newKubeResources. That way, when all the resources are being generated, it will already be able to consume the pull secret copy. WDYT? Does that make sense?
>
Ya I understand.

The challenge with doing everything before `newKubeResources` though is we need a deployment object that the `ensurePullSecret` function can modify.  A lot of what I came up with I found to be outside of what the `new.go` functions do.  Maybe you have some ideas.   

What I like about the current change is that even though I'm doing some additional resource regeneration, I'm doing it out in the open.  So I consider the kubeResources loop and post modify loop combine to make up the "resource generation loops", which is much more clear.

At the end of the day, we're talking about generating an object a few additional times.  So I guess I consider the code change to me mostly cosmetic.  Maybe I missing something.

mhrivnak: Rather than modifying the Deployment, why not just have the deployment always reference the secret from the beginning? I assume it would go like this:

1) ensure things exist that are based on non-CR inputs: `ensurePullSecret` and `ensureAdditionalTrustedCA`.
2) generate resource definitions assuming the above has been accomplished. The resources created in step 1 will be referenced by these definitions.
3) ensure the generated resources exist, include all other `ensure*` functions.

rthallisey: If I'm understanding you correctly, the deployment spec would look have this in it:

```yaml
                                                corev1.Volume{
                                                        Name: "cincinnati-graph-data",
                                                        VolumeSource: corev1.VolumeSource{
                                                                EmptyDir: &corev1.EmptyDirVolumeSource{},
                                                        },                                                                                                                                                                                                                                      
```

rthallisey: If I'm understanding you correctly, the deployment spec would have these by default:

```yaml
...

                                Spec: corev1.PodSpec{                                                                                                                                                                                                                                
                                         Volumes: []corev1.Volume{ 
                                             VolumeSource: corev1.VolumeSource{                                                                                                                                                                                                                              
                                                Secret: &corev1.SecretVolumeSource{                                                                                                                                                                                                                  
                                                    SecretName: NamePullSecret,                                                                                                                                                                                                                  
                                                    DefaultMode: &mode,                                                                                                                                                                                                                  
                                                },                                                                                                                                                                                                                                   
                                              },                                                                  
                                              VolumeSource: corev1.VolumeSource{                                                                                                                                                                                                                                              
                                                 ConfigMap: &corev1.ConfigMapVolumeSource{                                                                                                                                                                                                            
                                                     DefaultMode: &mode,                                                                                                                                                                                                                          
                                                     LocalObjectReference: corev1.LocalObjectReference{                                                                                                                                                                                                              
                                                         Name: nameAdditionalTrustedCA(instance),                                                                                                                                                                                            
                                                     },                                                                                                                                                                                                                                           
                                                     Items: []corev1.KeyToPath{                                                                                                                                                                                                                           
                                                        corev1.KeyToPath{                                                                                                                                                                                                                                    
                                                        Path: "tls-ca-bundle.pem",
                                                        Key:  NameCertConfigMapKey,                                                                                                                                                                                                          
                                                     },                                                                                                                                                                                                                                   
                                                 },                                                                                                                                                                                                                                   
                                              },                                                                                                                                                                                                                                   
                                          },        
                                      },                                                                                                                                                                                                     
```

`ensurePullSecret` runs to setup the pull-secret copy.  `ensureAdditionalTrustedCA` runs to setup the externalCA configmap copy.  Normal resource generation occurs and the remaining `ensure*` functions run.

I think this will work for the pull-secret because we should always have a copy.   However, it's fine to [not have a externalCA configmap](https://github.com/openshift/cincinnati-operator/blob/master/pkg/controller/cincinnati/cincinnati_controller.go#L233), so I don't think the same pattern works there.

mhrivnak: Can we just pass the configmap's name as an optional parameter to `newKubeResources`? Given the choice of telling it up-front that the configmap exists, or waiting until later to add it, I think it's preferable to do it up-front.

mhrivnak: Following some offline chat, I think this is ok as-is and don't see a need to block the PR.

My TL;DR: I think that `newKubeResources` is a constructor, and we should let it do all of the constructing, rather than defer some of the constructing to outside logic. This code as-is does make it clear where the deferred constructing should happen, which is helpful toward alleviating my concerns about leaking constructor logic outside of one well-known place. In this case, the logic is basically in two well-known places instead of one.

pkg/controller/cincinnati/new.go
```diff
@@ -333,36 +321,105 @@ func (k *kubeResources) newDeployment(instance *cv1beta1.Cincinnati) *appsv1.Dep
...
+		corev1.Volume{
+			Name: "cincinnati-graph-data",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
+	}
+}
+
+func (k *kubeResources) newTrustedCAConfig(instance *cv1beta1.Cincinnati, cm *corev1.ConfigMap) {
```
mhrivnak, line 347: By convention, a function that starts with `new` is a constructor that returns something new. Since this function is intended to have a side effect of modifying something that exists vs. returning something new, I suggest a name change to `addTrustedCAConfig` or something similar.

pkg/controller/cincinnati/new.go
```diff
@@ -333,36 +321,105 @@ func (k *kubeResources) newDeployment(instance *cv1beta1.Cincinnati) *appsv1.Dep
...
+	k.trustedCAConfig = &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nameAdditionalTrustedCA(instance),
+			Namespace: instance.Namespace,
+		},
+		Data: cm.Data,
+	}
+}
+
+func (k *kubeResources) newPullSecret(instance *cv1beta1.Cincinnati, s *corev1.Secret) {
```
mhrivnak, line 357: Same as above.

pkg/controller/cincinnati/new.go
```diff
@@ -333,36 +321,105 @@ func (k *kubeResources) newDeployment(instance *cv1beta1.Cincinnati) *appsv1.Dep
...
+	k.pullSecret = &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      namePullSecretCopy(instance),
+			Namespace: instance.Namespace,
+		},
+		Data: s.Data,
+	}
+}
+
+// regenerate - rebuild smaller resources to regerate the larger object
```
mhrivnak, line 367: s/regerate/regenerate/

pkg/controller/cincinnati/new.go
```diff
@@ -460,6 +507,20 @@ func (k *kubeResources) newGraphBuilderContainer(instance *cv1beta1.Cincinnati,
 	return g
 }
 
+func (k *kubeResources) newGraphBuilderVolumeMounts(instance *cv1beta1.Cincinnati) *[]corev1.VolumeMount {
```
mhrivnak, line 510: nit: I don't think there's an advantage here to returning the pointer to the slice. The slice already is essentially a reference.

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -214,39 +234,117 @@ func handleCACertStatus(reqLogger logr.Logger, status *cv1beta1.CincinnatiStatus
...
 		m := fmt.Sprintf("Found ConfigMap referenced by ImageConfig.Spec.AdditionalTrustedCA.Name but did not find key 'cincinnati-registry' for registry CA cert in ConfigMap (Name: %v, Namespace: %v)", image.Spec.AdditionalTrustedCA.Name, openshiftConfigNamespace)
 		handleCACertStatus(reqLogger, &instance.Status, "EnsureAdditionalTrustedCAFailed", m)
+		return nil, nil
+	}
+
+	return sourceCM, nil
+}
+
+func (r *ReconcileCincinnati) ensurePullSecret(ctx context.Context, reqLogger logr.Logger, instance *cv1beta1.Cincinnati, resources *kubeResources) error {
+	sourcePS, err := r.findPullSecret(ctx, reqLogger, instance, resources)
```
mhrivnak, line 323: Isn't this going to find the one in openshift's config namespace? I also see that it's not using this value later. Do we need to do this at all here?

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -214,39 +234,117 @@ func handleCACertStatus(reqLogger logr.Logger, status *cv1beta1.CincinnatiStatus
...
+	if err := r.ensureSecret(ctx, reqLogger, resources.pullSecret); err != nil {
+		handleErr(reqLogger, &instance.Status, "EnsureSecretFailed", err)
+		return err
+	}
+
+	return nil
+}
+
+func (r *ReconcileCincinnati) ensureAdditionalTrustedCA(ctx context.Context, reqLogger logr.Logger, instance *cv1beta1.Cincinnati, resources *kubeResources) error {
+	sourceCM, err := r.findTrustedCAConfig(ctx, reqLogger, instance, resources)
```
mhrivnak, line 344: Same as above, why call this here?

pkg/controller/cincinnati/names.go
```diff
@@ -18,6 +18,8 @@ const (
 	NameTrustedCAVolume = "trusted-ca"
 	// NameCertConfigMapKey is the ConfigMap key name where the operator expects the external registry CA Cert
 	NameCertConfigMapKey = "cincinnati-registry"
+	// NamePullSecret is the OpenShift pull secret name
+	NamePullSecret = "pull-secret"
```
mhrivnak, line 22: Since I don't think you're expecting this to get used outside of the `namePullSecretCopy` function, is there value in exporting it or even having it as a constant?

pkg/controller/cincinnati/new.go
mhrivnak: Same above in `newVolumes`

pkg/controller/cincinnati/new.go
```diff
@@ -65,18 +66,22 @@ name = "edge-add-remove"`
...
+	deployment               *appsv1.Deployment
+	graphBuilderContainer    *corev1.Container
+	graphDataInitContainer   *corev1.Container
+	policyEngineContainer    *corev1.Container
+	graphBuilderService      *corev1.Service
+	policyEngineService      *corev1.Service
+	policyEngineRoute        *routev1.Route
+	trustedCAConfig          *corev1.ConfigMap
+	pullSecret               *corev1.Secret
+	volumes                  *[]corev1.Volume
```
mhrivnak, line 83: What's your thinking on making these pointers to slices instead of just slices? Usually on a struct it's sufficient to reference a slice directly, since a slice itself is already very similar to a pointer.

pkg/controller/cincinnati/new.go
mhrivnak: Ah, I see you already have an `addTrustedCAConfig` function, and it's the only place where this gets called. What do you think about just moving this code into that function? I'm not seeing the value in having a separate function just to instantiate and assign a ConfigMap, and that would solve the naming confusion.

pkg/controller/cincinnati/new.go
mhrivnak: I'm generally not sure why this function is necessary. What is it about the container and the deployment that need to be regenerated?

pkg/controller/cincinnati/cincinnati_controller.go
rthallisey: Yes, this is all for error checking.  We need to answer the question:
    _Is there a non empty pull-secret in openshift-config?_
   **No** -> error.  There should always be a secret.
   **Yes** and **secret == nil** -> there's nothing to copy (user doesn't have registry auth).  I can see a case for this being an error.  However, this is a weird state that I don't know how the cluster is functional at all, so I'm inclined to leave it.
   **Yes** and **secret != nil** -> continue.  Ensure the pull secret is in the local ns

pkg/controller/cincinnati/cincinnati_controller.go
rthallisey: Similar to above except `findTrustedCAConfig` handles different scenarios.

Potential errors from `findTrustedCAConfig` are:
_Found image.config.openshift.io.Spec.AdditionalTrustedCA.Name but did not find expected ConfigMap (Name: %v, Namespace: %v)_
_Failed to GET the object_ (API timeout or error)

The question we're trying to answer here is:
_Were we able to find a non empty configmap named `image.config.openshift.io.Spec.AdditionalTrustedCA.Name` in openshift-config?_
**No there was an error** -> error
**Yes** and **Configmap == nil** -> return nil
**Yes** and **Configmap != nil** -> continue.  Ensure configmap.

pkg/controller/cincinnati/cincinnati_controller.go
mhrivnak: I'm not sure what you mean here by `secret == nil`. What value are you referring to exactly?

It seems like if reconciliation got this far, then it did find a pull secret in openshift-config, and it already used that to generate the desired state for a local copy of that secret. Once we get here, why do we need to re-check for the original? Why not just proceed with ensuring the desired state that was previously decided upon?

pkg/controller/cincinnati/new.go
rthallisey: Since we're modifying the graph builder container object and not the deployment, the deployment needs to be regenerated to pickup the changes.  I made two changes on that concept:
1) I pulled the regeneration logic out of `ensureAdditionalTrustedCA` to make it visible as the final step in the post resource modification process. 
2) I broke the objects I'm modifying down to their smallest form to more easily manipulate them.  Then, I rebuild each larger object one by one, which I found it to be less error prone since it only requires one call to `newDeployment`.

For example:
1) Append the pull-secret volumeMount to the graphBuilderVolumeMounts
2) Rebuild the graphBuilderContainer object
3) Rebuild the deployment

pkg/controller/cincinnati/cincinnati_controller.go
rthallisey: I'm assuming the GET can return an empty secret where as `secret == nil`.  

Right, we answered the question: _Is there a non empty pull-secret in openshift-config?_ already in `postAddPullSecret`. So the question is whether there is a point to checking it a second time.  The only value I see is if a user modified the pull-secret in the time between function calls.   I think that's pretty unlikely and we would _eventually_ catch it.

pkg/controller/cincinnati/new.go
rthallisey: That's fine with me.

pkg/controller/cincinnati/new.go
rthallisey: Just a mistake.  I'm using like it was `[]corev1.Volume`, so it should really just be that.

pkg/controller/cincinnati/names.go
rthallisey: I'm being a 🐑 here and exporting like every other constant.  Tbh, I think your question can be asked of all these vars.  But, it might be better to deal with this in a follow up PR.

mhrivnak: I'll take a stab at answering that now. :) None of the other exported constants here are used as part of a resource name. They are all identifiers used within a resource, such as the name of a container of volume. And if you look at the functions below that are generating names, none of them are using exported constants.

The container names are useful outside this package, and probably the volume name is too. For example, they are what someone would use on the command line to run `oc logs -c policy-engine $PODNAME`. They don't change from one operand instance to another. That's why I think they're useful to export, which implicitly documents them.

How resource names are generated is so far not part of this operator's public API, whether the golang API or k8s API. So I think it would be best to not export this.

pkg/controller/cincinnati/cincinnati_controller.go
```diff
@@ -120,6 +119,13 @@ type ReconcileCincinnati struct {
 // Reconcile reads that state of the cluster for a Cincinnati object and makes changes based on the state read
 // and what is in the Cincinnati.Spec
 func (r *ReconcileCincinnati) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	/*    **Reconcile Pattern**
+	1. Create all the kubeResources
+	2. Make a few modifications to the kubeResources
+	3. Regenerate some of the kubeResources
+	4. Ensure all the kubeResources are correct in the Cluster
+	*/
```
djzager, line 127: Personally, I think we should drop this comment in favor of more detailed comments where appropriate. For example, where we loop over `postAddPullSecret` and `postAddExternalCACert` would be a good place to specifically mention that we are updating our reference `resources` to add a pull secret and external CA cert if found.

rthallisey: Added some more detailed comments, but I kept this section to use as an index.  Let me know what you think.

PR #55, opened 2020-08-04
pkg/apis/cincinnati/v1beta1: Explicitly require registry and repository

wking:
Users must point the update service at a repository so the graph builder can scrape for primary release metadata.

Also require the `spec`, but leave `status` optional.  This allows users to upload a manifest, while leaving the status empty for the controller to fill in later.

WIP because I need to figure out how to regenerate the machine-written portions of this package.


wking:
No longer a WIP with b715a56 -> 2347cb1.  Also made the other spec properties required for now.  Discussion in the 2347cb1 commit message.

PR #56, opened 2020-08-04
vendor/github.com/openshift/build-machinery-go: Pull in generic helpers

wking:
Generated by writing the `Makefile` and `pkg/dependencymagnet/doc.go` changes by hand, and then running:

```console
$ go mod tidy
$ go mod vendor
```

with:

```console
$ go version
go version go1.14.4 linux/amd64
```

I've also removed a bunch of auto-generated editor stuff from `.gitignore`.  From [the docs][1]:

> Which file to place a pattern in depends on how the pattern is meant to be used.
>
> * Patterns which should be version-controlled and distributed to other repositories via clone (i.e., files that all developers will want to ignore) should go into a `.gitignore` file.
> * ...
> * Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user's editor of choice) generally go into a file specified by `core.excludesFile` in the user's `~/.gitconfig`. Its default value is `$XDG_CONFIG_HOME/git/ignore`. If `$XDG_CONFIG_HOME` is either not set or empty, `$HOME/.config/git/ignore` is used instead.

So editor files are not appropriate for the project's versioned-controlled `.gitignore`.  We could keep the Go stuff, but it's not clear to me if we have tooling that generates any of those outside of the temporary build directories which already have manually-curated `.gitignore` entries.

[1]: https://git-scm.com/docs/gitignore#_description


wking:
Hmm, `make update-generated` does not seem to be regenerating `pkg/apis/cincinnati/v1beta1/zz*` with this change.  Not sure why not yet.

/hold

wking:
/close

Obsoleting in favor of #59, to stick to the operator-sdk tooling.

PR #57, opened 2020-08-04
pkg/apis/cincinnati/v1beta1: Drop redundant doc.go

wking:
We've had a package docstring in both doc.go and register.go since v1alpha1 in 52a976598c (#1).  But there's no reason to have a `doc.go` declaration when `register.go` already defines the same material.

Also add a trailing period to the docstring sentence and remove the boilerplate comment, because this is a hand-maintained file, even if changes will be rare.


wking:
#65 will clear this up as a side-effect.

/close

PR #59, opened 2020-08-04
README: Document operator-sdk installation

wking:
So new folks understand where the `operator-sdk` command should come from, or remember to look in `vendor/` to see which version to use:

```console
$ grep '#.*operator-sdk' vendor/modules.txt 
```


LalatenduMohanty (approved):
/lgtm

PR #61, opened 2020-08-10
Added disconnected cincinnati docs

mvazquezc:
Added docs for:
- Deploying Cincinnati operator on a disconnected OCP cluster
- Using Cincinnati operator for updating a disconnected OCP cluster


mhrivnak:
The install process described here is different from what [OCP docs contain](https://docs.openshift.com/container-platform/4.5/operators/olm-restricted-networks.html). Why the difference?

mvazquezc:
> The install process described here is different from what [OCP docs contain](https://docs.openshift.com/container-platform/4.5/operators/olm-restricted-networks.html). Why the difference?

As far as I know, the cincinnati-operator is not published in the default OpenShift Operator Catalogs yet. Thus, following the doc you linked wouldn't make the operator available for the disconnected cluster. This doc explains how to build a custom catalog and add it to the disconnected cluster.

When the cincinnati-operator is part of a default catalog (Red Hat Operators?) then we should change the docs and add that to the doc.

mhrivnak:
> > The install process described here is different from what [OCP docs contain](https://docs.openshift.com/container-platform/4.5/operators/olm-restricted-networks.html). Why the difference?
> 
> As far as I know, the cincinnati-operator is not published in the default OpenShift Operator Catalogs yet. Thus, following the doc you linked wouldn't make the operator available for the disconnected cluster. This doc explains how to build a custom catalog and add it to the disconnected cluster.
> 
> When the cincinnati-operator is part of a default catalog (Red Hat Operators?) then we should change the docs and add that to the doc.

Sounds good as a workaround. I suggest adding that qualifier prominently to this doc in case someone finds this in the future, to ensure we don't confuse anyone.

mvazquezc:
> > > The install process described here is different from what [OCP docs contain](https://docs.openshift.com/container-platform/4.5/operators/olm-restricted-networks.html). Why the difference?
> > 
> > 
> > As far as I know, the cincinnati-operator is not published in the default OpenShift Operator Catalogs yet. Thus, following the doc you linked wouldn't make the operator available for the disconnected cluster. This doc explains how to build a custom catalog and add it to the disconnected cluster.
> > When the cincinnati-operator is part of a default catalog (Red Hat Operators?) then we should change the docs and add that to the doc.
> 
> Sounds good as a workaround. I suggest adding that qualifier prominently to this doc in case someone finds this in the future, to ensure we don't confuse anyone.

Added a disclaimer, let me know if that's enough.

mhrivnak:
That's great. Thanks!

mvazquezc:
Can this be merged?

PR #64, opened 2020-09-15
pkg/controller/cincinnati: Use InsecureEdgeTerminationPolicyNone

wking:
Builds on #63; you may want to review that first.

We had used `InsecureEdgeTerminationPolicyAllow` since the route landed in 1fdf865825 (#30).  The motivation for that value didn't make it into the Git commit message, but from [discussion in the GitHub pull request][1], it was:

* `InsecureEdgeTerminationPolicyAllow` is the default termination policy.
* [Cincinnati's docs][2] have no preference.

However, we really, really want HTTPS security for cluster-version operators making upstream requests for update recommendations.  There are long-term plans for tightening down guards against malicious, compromised, or man-in-the-middled update recommendation services, but today we have yet to land even guards as basic as "[upstream is lying about the version string associated with a given release image][3]".

By [removing HTTP termination][4], we force consumers to configure their clients, including the cluster-version operator, with `https://` URIs (or do something else explicit like setting up their own HTTP termination) before they can access the policy-engine output, which reduces the risk that they will recieve and trust compromised update graphs.  This may be a breaking change, but:

* We're still in beta, and not yet in general-availability with backwards-compatability requirements.
* Folks who have configured their cluster-version operators and other clients with `http://` upstreams should *want* to be broken.  We are protecting them from all sorts of compromised-upstream failure modes.
* The cluster-version operator, and other well-behaved clients, will report understandable error messages for "I tried to connect over HTTP and there was nobody there", which will lead users into auditing and fixing their upstream URIs, so recovering from the breakage should not be to onerous.

[1]: https://github.com/openshift/cincinnati-operator/pull/30#discussion_r418741486
[2]: https://github.com/openshift/cincinnati/blame/0bb5f6f3228858f9e5d1807bd6f45f46e537cdea/docs/user/running-cincinnati.md#L87-L88
[3]: https://github.com/openshift/cluster-version-operator/pull/431
[4]: https://github.com/openshift/api/blob/346618ed7d5e6396191efe6f10b2c36f1e95d8b7/route/v1/types.go#L258-L259

### DIFF
--- a/docs/disconnected-cincinnati-operator.md
+++ b/docs/disconnected-cincinnati-operator.md
@@ -237,7 +237,7 @@ You might want to review the documentation around disconnected registries to lea
 3. Check the cincinnati service
 
     ~~~sh
-    curl --header 'Accept:application/json' http://$(oc -n "${NAMESPACE}" get route example-name-policy-engine-route -o jsonpath='{.spec.host}')/api/upgrades_info/v1/graph\?channel=stable-4.5 | jq
+    curl --header 'Accept:application/json' https://$(oc -n "${NAMESPACE}" get route example-name-policy-engine-route -o jsonpath='{.spec.host}')/api/upgrades_info/v1/graph\?channel=stable-4.5 | jq
     ~~~
 
     > **OUTPUT**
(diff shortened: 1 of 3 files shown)


jottofar:
Tested the https (without configuring a cert) aspect of this change and appears okay.

/lgtm

wking:
I dunno what was wrong with [the deployment][1].  Maybe it was just a flake...

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-operator/64/pull-ci-openshift-cincinnati-operator-master-cincinnati-operator-e2e-aws/1305949506075889664

PR #72, opened 2020-10-23
functests: Ignore metrics for now

wking:
These aren't showing up, but I don't care about metrics at the moment.  Will revisit if we can get green CI.


wking:
/close

Issue was openshift/release#13110

PR #73, opened 2020-10-23
Dockerfile: Shift to the repo root

wking:
No need to shift this down into a subdirectory.  Add a symlink to give existing consumers some time to migrate.


wking:
Image got built.

PR #65, opened 2020-09-30
Migrate OSUS operator from v0.18.2 to v1.0

jottofar:

PR #69, opened 2020-10-19
functests: Drop Ginkgo dependency

wking:
Use Go's built-in testing package instead.  This avoids issues [like][1]:

```
Retrieving ginkgo and gomega build dependencies
go: github.com/openshift/cluster-image-registry-operator@v0.0.0-20200415091009-99c06ee64540 requires
		github.com/openshift/library-go@v0.0.0-20200327125526-163b2f0d6264 requires
		bitbucket.org/ww/goautoneg@v0.0.0-20120707110453-75cd24fc2f2c: reading https://api.bitbucket.org/2.0/repositories/ww/goautoneg?fields=scm: 404 Not Found
go: github.com/openshift/cluster-image-registry-operator@v0.0.0-20200415091009-99c06ee64540 requires
		github.com/openshift/library-go@v0.0.0-20200327125526-163b2f0d6264 requires
		bitbucket.org/ww/goautoneg@v0.0.0-20120707110453-75cd24fc2f2c: reading https://api.bitbucket.org/2.0/repositories/ww/goautoneg?fields=scm: 404 Not Found
hack/functest.sh: line 17: /go/bin/ginkgo: No such file or directory
hack/functest.sh: line 19: /go/bin/ginkgo: No such file or directory
```

The tests are simple enough that Ginkgo structuring was overkill anyway.  Generated by manually editing files in `functests`, and then running:

```console
$ go mod tidy
$ go mod vendor
$ git add -A go.* vendor
```

using:

```console
$ go version
go version go1.15.2 linux/amd64
```

[1]: https://github.com/openshift/cincinnati-operator/pull/66#issuecomment-703931608

PR #74, opened 2020-10-26
functests: Add set up method to add schemes

wking:
Pulled out of #65, since this should green up CI and seems orthogonal to the SDK pivot.

### DIFF
--- a/functests/cincinnati_creation_test.go
+++ b/functests/cincinnati_creation_test.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"os"
 	"testing"
 
 	routev1 "github.com/openshift/api/route/v1"
@@ -13,8 +14,24 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/klog"
 	client "sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+// TestMain will perform test setup, exec each test, and perform test tear down.
+func TestMain(m *testing.M) {
+	setUp()
+	retCode := m.Run()
+	os.Exit(retCode)
+}
+
+// setUp performs test setup.
+func setUp() {
+	if err := routev1.AddToScheme(scheme.Scheme); err != nil {
+		klog.Fatalf("Failed adding route to scheme, %v", err)
+	}
+}
 
 func TestCustomResource(t *testing.T) {
 	k8sClient, err := getK8sClient()


wking:
[Install][1]:

```
Error 400: Service account ci-op-r9q361-openshift-i-9ql9w@openshift-gce-devel-ci.iam.gserviceaccount.com does not exist., badRequest
```

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-operator/74/pull-ci-openshift-cincinnati-operator-master-operator-e2e/1320880902997282816

wking:
[e2e][1]:

```
 level=fatal msg=failed to fetch Cluster: failed to fetch dependency of "Cluster": failed to generate asset "Platform Provisioning Check": metadata.name: Invalid value: "ci-op-rd01z23c-fb094": record api.ci-op-rd01z23c-fb094.origin-ci-int-gce.dev.openshift.com. already exists in DNS Zone (openshift-gce-devel-ci/origin-ci-int-gce-new) and might be in use by another cluster, please remove it to continue
```

```console
$ oc --context build01 delete project ci-op-nhr8xycy
```

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-operator/74/pull-ci-openshift-cincinnati-operator-master-operator-e2e/1320917151908892672

wking:
All green :).

/assign @jottofar

functests/cincinnati_creation_test.go
```diff
@@ -13,9 +14,25 @@ import (
...
+// TestMain will perform test setup, exec each test, and perform test tear down.
+func TestMain(m *testing.M) {
+	setUp()
+	retCode := m.Run()
+	os.Exit(retCode)
+}
+
+// setUp performs test setup.
+func setUp() {
+	if err := routev1.AddToScheme(scheme.Scheme); err != nil {
```
jottofar, line 31: Is there a reason you dropped this?

https://github.com/openshift/cincinnati-operator/blob/2df239a8486d2ba3aa0d9925e5d505105ab36afe/functests/functests_suite_test.go#L16

wking: Yeah, we don't import that yet.  We can add it with the commit that adds the import

jottofar: But you're just adding back stuff that shouldn't have been removed to begin with. I should have caught it and commented in in the `Remove Ginkgp` commit so not understanding why this commit doen't just add it all back in.

wking: We [used to][1] add the route scheme (twice?).  And out local APIs, which you link.  I dunno why we don't actually need those schemes injected, but tests are green, so we must not need them.

[1]: https://github.com/openshift/cincinnati-operator/pull/69/files#diff-30ccb8be4f6b0bb46f097a71158dd8e4b22b959348fee463959aae62fb041d29L20

PR #75, opened 2020-10-28
Makefile: Inline the functest invocation

wking:
And drop the one-line functest.sh wrapper.  Also add a `-v` so we get the `t.Log` output to help debug failures.


LalatenduMohanty (approved):
/lgtm

PR #76, opened 2020-11-10
*: Scaffold for 0.19.3

wking:
Part of #65, but split up to make the fresh 0.19 scaffold a separate commit that gets merged into the main branch.  This makes it easier to see which lines were written by SDK tooling, which were inherited from master, and which are new from the merge.
We don't want to commit bin, so the warning is fine.


wking:

/close

PR #78, opened 2020-11-10
*: Pivot from v1beta1 Cincinnati to v1 UpdateService

wking:
Builds on #76.

Similar to #65's d31969ef0fe, but via re-generating with the SDK tooling to distinguish those changes from the ones we manually apply afterwards.


wking:

/close

PR #79, opened 2020-11-11
*: Scaffold for 1.0.1

wking:
Building on #78.  With this, we get reasonably close to #65.  I'll leave comments in #65 talking over any divergence.

PR #80, opened 2020-11-12
v1 scaffolding, but without the merges

wking:
Same content as #79, but without the merges that @LalatenduMohanty didn't like ;)


wking:

/close

PR #81, opened 2020-11-12
Changing the updateservice name to update-service

LalatenduMohanty:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,11 @@
 	build
 
 SOURCES := $(shell find . -name '*.go' -not -path "*/vendor/*")
-IMG ?= quay.io/updateservice/updateservice-operator:latest
+
+# This is a placeholder for cincinnati-operator image placeholder
+# During development override this when you want to use an specific image
+# Example: IMG ?= quay.io/jottofar/updateservice-operator-index:v1
+IMG ?= controller:latest
 
 clean:
 	@echo "Cleaning previous outputs"
@@ -25,4 +29,4 @@ unit-test:
 	go test -v ./controllers/...
 
 build: $(SOURCES)
-	go build -i -ldflags="-s -w" -mod=vendor -o ./updateservice-operator ./
+	go build -i -ldflags="-s -w" -mod=vendor -o ./update-service-operator ./
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# updateservice-operator
+# update-service-operator
 
 This operator is developed using the [operator SDK][operator-sdk], version 1.0.0.
 Installation docs are [here][operator-sdk-installation].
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -6,7 +6,7 @@
 	build
 
 SOURCES := $(shell find . -name '*.go' -not -path "*/vendor/*")
-IMG ?= quay.io/updateservice/updateservice-operator:latest
+IMG ?= quay.io/updateservice/update-service-operator:latest
```
wking, line 9: do we actually control this Quay namespace?  I prefer [the SDK's scaffolded `IMG ?= controller:latest`][1], which is more obviously a placeholder.

[1]: https://github.com/openshift/cincinnati-operator/commit/cfc695634ec34a3c602f34176fea7215fe3e73a3#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R15

README.md
```diff
@@ -15,18 +15,18 @@ operator-sdk run --local
 ## Using an init container to load graph data
 
 The UpdateService graph data is loaded from an init container. Before deploying 
-the updateservice-operator, you will need to [build and push an init container containing the graph data](docs/graph-data-init-container.md).
+the update-service-operator, you will need to [build and push an init container containing the graph data](docs/graph-data-init-container.md).
```
wking, line 18: can we just say "the operator" here?

LalatenduMohanty: Yeah, agree

Makefile
jottofar: I'm not sure how this works either but there's an existing `cincinnati` organization, quay.io/cincinnati, containing repos cincinnati and cincinnati-operator. So I thought we would have the same but with update-service.

jottofar: Also, whatever we decide, this default is also used other places, most notably in `config/manager/manager.yaml`, which then allows use of `kustomize`,  or `hack/deploy.sh`, to change it to a chosen image.

LalatenduMohanty: Right now I have used `IMG ?= controller:latest` in all the places as we do not have `quay.io/updateservice/update-service-operator` in quay. Once we figure out what namespace we want to use for the CI we can change these back.

LalatenduMohanty: Resolved in https://github.com/openshift/cincinnati-operator/commit/f99ff7df60ccd1ffae42fcb1a4cd8b6f3d78027f

README.md
LalatenduMohanty: fixed in https://github.com/openshift/cincinnati-operator/commit/f99ff7df60ccd1ffae42fcb1a4cd8b6f3d78027f

README.md
```diff
@@ -8,25 +8,25 @@ Installation docs are [here][operator-sdk-installation].
 To run locally, you must set the operand image as shown below.
 
 ```
-export OPERAND_IMAGE="quay.io/app-sre/updateservice:2873c6b" 
+export OPERAND_IMAGE="quay.io/app-sre/update-service:2873c6b"
```
jottofar, line 11: I should have left the image name as cincinnati when I did the renaming. I think it best to change it back to `quay.io/app-sre/cincinnati:2873c6b` to match `hack/deploy.sh` until we know what the updated OSUS image will be to use as a default.

LalatenduMohanty: I agree.

LalatenduMohanty: fixed https://github.com/openshift/cincinnati-operator/commit/2f41139e45ed957fcb2062cf8e7a7b64df53f61b

PR #62, opened 2020-08-26
Add OTA team as OWNERS

rthallisey:
@LalatenduMohanty  @wking @steveeJ @jottofar @vrutkovs

PR #89, opened 2020-11-18
OWNERS: Drop Advanced Cluster Management folks

wking:
The ACM devs kindly scaffolded this operator, but have since handed it off to the updates folks.  This commit completes the ownership transition begun in f7860a3b9b (#62).

I've also applied a case-insensitve sort to the remaining entries.


wking:

No need to run CI on an `OWNERS` change.

PR #90, opened 2020-11-18
OWNERS: Change 'component' to 'product'

wking:
I'd updated the value in 727a3c58f8 (#77), but at that point I was uncertain about how to handle Bugzilla products vs. components.  Today, Luke pointed me towards [the consuming code][1], showing that we should have used the 'product' key, overriding ART's `OpenShift Container Platform` default product.

[1]: https://github.com/openshift/doozer/blob/231d240702a5c22da488fa241bc76df416acee6a/doozerlib/metadata.py#L288-L343

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -13,4 +13,4 @@ reviewers:
 - vrutkovs
 - wking
 
-component: OpenShift Update Service
+product: OpenShift Update Service


wking:
Rebased, now that #89 has landed.  No need to run CI on an OWNERS change.

LalatenduMohanty:
-1, as OSUS and OSUS is a sub-product is under OCP.

wking:
>  OSUS and OSUS-operator is a sub-product is under OCP.

There are no sub-products in Bugzilla, which is what this config entry is feeding.  And the `curl` [here][1] shows that we are our own Bugzilla product today.  But sounds like we have plans to move back under the OCP product and become a Bugzilla component again, so:

/close

[1]: https://github.com/openshift/cincinnati-operator/pull/77/commits/727a3c58f82581f60dd444a376fd1cf0d9013407

PR #93, opened 2020-11-30
Operator bundle: changes to support operator bundling

jottofar:
Modifies Makefile and creates the config/manifests artifacts which support bundling and is generated by the new recipe 'gen-for-bundle'.

In its current form this PR assumes that the official build will run the `bundle` target and therefore only needs the `config/manifests` directory. Another possibility is that the official build only builds the bundle image in which case we would run and check-in the results of the `bundle` recipe. The official build would then use these artifacts to build the bundle image which is then used to create or update a bundle index.

The additional recipes added to the Makefile are described in detail below.

`manifests` generates manifests, e.g. CRD, RBAC etc., and is run by development when required.

`gen-for-bundle` generates kustomize bases and a kustomization.yaml used to generate the operator bundle. This was originally part of the `bundle` recipe as generated by the operator-sdk scaffolding. I have broken it out under the assumption that we would run this recipe and check-in the resulting config/manifests directory and contents to the git repo where they would be used by official build tools which actually generate the operator bundle.

`bundle` generates the bundle and metadata as contained in a `bundle` directory and bundle.Dockerfile. A `BUNDLE_VERSION` variable has been added to allow setting bundle version. If it is determined that this is to be ran by development, and not ran as part of the official build, the results of this step should also be checked into git to be used by the official build tools.

`bundle-build` builds the bundle image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,11 @@ VERSION ?= 0.0.1
 	deploy \
 	func-test \
 	unit-test \
-	build
+	build \
+	manifests \
+	gen-for-bundle \
+	bundle \
+	bundle-build
 
 SOURCES := $(shell find . -name '*.go' -not -path "*/vendor/*")
 GOBUILDFLAGS ?= -i -mod=vendor
@@ -14,8 +18,21 @@ GOLDFLAGS ?= -s -w -X github.com/openshift/cincinnati-operator/version.Operator=
 
 # This is a placeholder for cincinnati-operator image placeholder
 # During development override this when you want to use an specific image
-# Example: IMG ?= quay.io/jottofar/updateservice-operator-index:v1
+# Example: IMG ?= quay.io/jottofar/update-service-operator:v1
 IMG ?= controller:latest
+
+BUNDLE_IMG ?= controller-bundle:latest
+
+CHANNELS ?= v1
+BUNDLE_CHANNELS = --channels=$(CHANNELS)
+
+DEFAULT_CHANNEL ?= v1
+BUNDLE_DEFAULT_CHANNEL = --default-channel=$(DEFAULT_CHANNEL)
+
+BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
+
+# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
+CRD_OPTIONS ?= "crd:trivialVersions=true"
 
 clean:
 	@echo "Cleaning previous outputs"
@@ -35,3 +52,53 @@ unit-test:
 
 build: $(SOURCES)
 	go build $(GOBUILDFLAGS) -ldflags="$(GOLDFLAGS)" -o ./update-service-operator ./
+
+# find or download controller-gen
+# download controller-gen if necessary
+controller-gen:
+ifeq (, $(shell which controller-gen))
+	@{ \
+	set -e ;\
+	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$CONTROLLER_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0 ;\
+	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
+	}
+CONTROLLER_GEN=$(GOBIN)/controller-gen
+else
+CONTROLLER_GEN=$(shell which controller-gen)
+endif
+
+kustomize:
+ifeq (, $(shell which kustomize))
+	@{ \
+	set -e ;\
+	KUSTOMIZE_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$KUSTOMIZE_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/kustomize/kustomize/v3@v3.5.4 ;\
+	rm -rf $$KUSTOMIZE_GEN_TMP_DIR ;\
+	}
+KUSTOMIZE=$(GOBIN)/kustomize
+else
+KUSTOMIZE=$(shell which kustomize)
+endif
+
+# Generate manifests e.g. CRD, RBAC etc.
+manifests:  controller-gen
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=updateservice-operator webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+
+# Generate bundle manifests
+gen-for-bundle: manifests kustomize
+	operator-sdk generate kustomize manifests -q
+
+# Generate bundle and metadata, then validate generated files.
+bundle: kustomize
+	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
+	operator-sdk bundle validate ./bundle
+
+# Build the bundle image.
+bundle-build:
+	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
--- a/config/manifests/bases/update-service-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/update-service-operator.clusterserviceversion.yaml
@@ -1,0 +1,90 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    alm-examples: '[]'
+    capabilities: Basic Install
+    description: Creates and maintains an OpenShift Update Service instance
+    operatorframework.io/suggested-namespace: openshift-update-service
+    operators.operatorframework.io/builder: operator-sdk-v1.2.0
+    operators.operatorframework.io/project_layout: go.kubebuilder.io/v2
+  name: update-service-operator.vX.Y.Z
+  namespace: placeholder
+spec:
+  apiservicedefinitions: {}
+  customresourcedefinitions:
+    owned:
+    - description: UpdateService is the Schema for the updateservices API.
+      displayName: Update Service
+      kind: UpdateService
+      name: updateservices.updateservice.operator.openshift.io
+      version: v1
+  description: |-
+    # Use Case
+    Running an Update Service instance in a cluster is appealing for offline OpenShift
+    clusters or for admins that want to provide their own graph data instead of
+    using [https://github.com/openshift/cincinnati-graph-data](https://github.com/openshift/cincinnati-graph-data/).
+
+    # About Update Service
+    Update Service uses **Cincinnati** which is an update protocol designed to facilitate
+    automatic updates. It describes a particular method for representing transitions between
+    releases of a project and allowing a client to perform automatic updates between these
+    releases.
+
+    ## Components
+    A **Graph Builder** iterates over the release payloads hosted by the
+    storage component and builds a DAG of the releases. It is responsible for
+    verifying that the graph described by the releases is acyclic and connected.
+
+    A **Policy Engine** is in charge of altering a client's view of the graph
+    by applying a set of filters which are defined within the particular Policy
+    Engine instance. Both the input to and the output from Policy Engines is a
+    graph, allowing multiple Policy Engines to be chained together. The first
+    Policy Engine in a chain will fetch its graph from the Graph Builder and the
+    last Policy Engine in a chain will serve the modified graph to the client.
+
+    An **Update Service client** is the end consumer of the release payloads. The
+    client periodically queries the Policy Engine for updates and applys them if
+    available.
+
+    # Query OpenShift's Update Service Endpoint
+    $ curl --silent --header 'Accept:application/json' 'https://api.openshift.com/api/upgrades_info/v1/graph?arch=amd64&channel=stable-4.2' | jq '. as $graph | $graph.nodes | map(.version == "4.2.13") | index(true) as $orig | $graph.edges | map(select(.[0] == $orig)[1]) | map($graph.nodes[.])'
+  displayName: OpenShift Update Service
+  icon:
+  - base64data: PHN2ZyBpZD0iTGF5ZXJfMSIgZGF0YS1uYW1lPSJMYXllciAxIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAxOTIgMTQ1Ij48ZGVmcz48c3R5bGU+LmNscy0xe2ZpbGw6I2UwMDt9PC9zdHlsZT48L2RlZnM+PHRpdGxlPlJlZEhhdC1Mb2dvLUhhdC1Db2xvcjwvdGl0bGU+PHBhdGggZD0iTTE1Ny43Nyw2Mi42MWExNCwxNCwwLDAsMSwuMzEsMy40MmMwLDE0Ljg4LTE4LjEsMTcuNDYtMzAuNjEsMTcuNDZDNzguODMsODMuNDksNDIuNTMsNTMuMjYsNDIuNTMsNDRhNi40Myw2LjQzLDAsMCwxLC4yMi0xLjk0bC0zLjY2LDkuMDZhMTguNDUsMTguNDUsMCwwLDAtMS41MSw3LjMzYzAsMTguMTEsNDEsNDUuNDgsODcuNzQsNDUuNDgsMjAuNjksMCwzNi40My03Ljc2LDM2LjQzLTIxLjc3LDAtMS4wOCwwLTEuOTQtMS43My0xMC4xM1oiLz48cGF0aCBjbGFzcz0iY2xzLTEiIGQ9Ik0xMjcuNDcsODMuNDljMTIuNTEsMCwzMC42MS0yLjU4LDMwLjYxLTE3LjQ2YTE0LDE0LDAsMCwwLS4zMS0zLjQybC03LjQ1LTMyLjM2Yy0xLjcyLTcuMTItMy4yMy0xMC4zNS0xNS43My0xNi42QzEyNC44OSw4LjY5LDEwMy43Ni41LDk3LjUxLjUsOTEuNjkuNSw5MCw4LDgzLjA2LDhjLTYuNjgsMC0xMS42NC01LjYtMTcuODktNS42LTYsMC05LjkxLDQuMDktMTIuOTMsMTIuNSwwLDAtOC40MSwyMy43Mi05LjQ5LDI3LjE2QTYuNDMsNi40MywwLDAsMCw0Mi41Myw0NGMwLDkuMjIsMzYuMywzOS40NSw4NC45NCwzOS40NU0xNjAsNzIuMDdjMS43Myw4LjE5LDEuNzMsOS4wNSwxLjczLDEwLjEzLDAsMTQtMTUuNzQsMjEuNzctMzYuNDMsMjEuNzdDNzguNTQsMTA0LDM3LjU4LDc2LjYsMzcuNTgsNTguNDlhMTguNDUsMTguNDUsMCwwLDEsMS41MS03LjMzQzIyLjI3LDUyLC41LDU1LC41LDc0LjIyYzAsMzEuNDgsNzQuNTksNzAuMjgsMTMzLjY1LDcwLjI4LDQ1LjI4LDAsNTYuNy0yMC40OCw1Ni43LTM2LjY1LDAtMTIuNzItMTEtMjcuMTYtMzAuODMtMzUuNzgiLz48L3N2Zz4=
+    mediatype: image/svg+xml
+  install:
+    spec:
+      deployments: null
+    strategy: ""
+  installModes:
+  - supported: true
+    type: OwnNamespace
+  - supported: true
+    type: SingleNamespace
+  - supported: false
+    type: MultiNamespace
+  - supported: true
+    type: AllNamespaces
+  keywords:
+  - Update Service
+  - UpdateService
+  - update-service
+  - OSUS
+  - Cincinnati
+  - version
+  - upgrade
+  links:
+  - name: Update Service Operator
+    url: https://github.com/openshift/cincinnati-operator
+  - name: OpenShift Update Service
+    url: https://github.com/openshift/cincinnati
+  - name: OpenShift Graph Data
+    url: https://github.com/openshift/cincinnati-graph-data
+  maintainers:
+  - email: aos-team-ota@redhat.com
+    name: OpenShift Update Service maintainers
+  maturity: alpha
+  provider:
+    name: Red Hat
+  version: 0.0.0
--- a/config/manifests/kustomization.yaml
+++ b/config/manifests/kustomization.yaml
@@ -1,0 +1,4 @@
+resources:
+- ../default
+- ../samples
+- ../scorecard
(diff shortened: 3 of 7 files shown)


Makefile
```diff
@@ -1,22 +1,42 @@
 # Current Operator version
 VERSION ?= 0.0.1
 
+# Current bundle version
+BUNDLE_VERSION ?= 1.0.0
```
wking, line 5: SDK scaffolding [recycles `VERSION` for this][1].  I don't think we need a separate variable.  It's going to be set at build-time anyway, and we can always set it differently when building the bundle and operator if we feel a need to have divergent values.

[1]: https://github.com/openshift/cincinnati-operator/blob/0731f08b2fbd3989b4642ff55c6a216e97d4c990/Makefile#L4

Makefile
```diff
@@ -1,22 +1,42 @@
...
 # This is a placeholder for cincinnati-operator image placeholder
 # During development override this when you want to use an specific image
-# Example: IMG ?= quay.io/jottofar/updateservice-operator-index:v1
+# Example: IMG ?= quay.io/jottofar/update-service-operator:v1
 IMG ?= controller:latest
 
+BUNDLE_IMG ?= controller-bundle:latest
+
+CHANNELS ?= v1
+BUNDLE_CHANNELS = --channels=$(CHANNELS)
```
wking, line 30: SDK scaffolding [has some `ifdefs` around `BUNDLE_CHANNELS` and `BUNDLE_DEFAULT_CHANNEL`][1].  Do we know what those are about?  They also don't seem to have a default for `CHANNELS`.  Do we understand the flow for that?  Is it something that's usually set explicitly at build time?

[1]: https://github.com/openshift/cincinnati-operator/blob/0731f08b2fbd3989b4642ff55c6a216e97d4c990/Makefile#L6-L12

Makefile
```diff
@@ -35,3 +55,21 @@ unit-test:
 
 build: $(SOURCES)
 	go build $(GOBUILDFLAGS) -ldflags="$(GOLDFLAGS)" -o ./update-service-operator ./
+
+# Generate manifests e.g. CRD, RBAC etc.
+manifests:
+	controller-gen $(CRD_OPTIONS) rbac:roleName=updateservice-operator webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
wking, line 61: SDK scaffolding [includes a section on installing `controller-gen`][1], setting up a variable to use here.  Can we follow that approach?  I like having the expected version of `controller-gen` mentioned in version control.

[1]: https://github.com/openshift/cincinnati-operator/blob/0731f08b2fbd3989b4642ff55c6a216e97d4c990/Makefile#L77-L92
[2]: https://github.com/openshift/cincinnati-operator/blob/0731f08b2fbd3989b4642ff55c6a216e97d4c990/Makefile#L55

Makefile
```diff
@@ -35,3 +55,21 @@ unit-test:
...
+
+# Generate bundle manifests
+gen-for-bundle:
+	operator-sdk generate kustomize manifests -q
```
wking, line 65: From [your topic post][1]:

> `gen-for-bundle` generates kustomize bases and a kustomization.yaml used to generate the operator bundle. This was originally part of the `bundle` recipe as generated by the operator-sdk scaffolding. I have broken it out under the assumption that we would run this recipe and check-in the resulting config/manifests directory and contents to the git repo where they would be used by official build tools which actually generate the operator bundle.

I'm fine with that if we have access to `kustomize` and `operator-sdk` within the build environment, otherwise all of `bundle` has to be run by developers and committed to Git, right?  I don't see a point in diverging from [the scaffolded recipe][2] unless we can defer some of the logic to the build environment.

[1]: https://github.com/openshift/cincinnati-operator/pull/93#issue-529862615
[2]: https://github.com/openshift/cincinnati-operator/blob/0731f08b2fbd3989b4642ff55c6a216e97d4c990/Makefile#L109-L115

Makefile
```diff
@@ -35,3 +55,21 @@ unit-test:
...
+
+# Generate bundle and metadata, then validate generated files.
+bundle:
```
wking, line 68: SDK's scaffolded `bundle` [depends on `manifests`][1].  With your current split, I'd expect `bundle` to depend on `gen-for-bundle`, and then either `bundle` or `gen-for-bundle` (I'm not sure which) to depend on `manifests`.

[1]: https://github.com/openshift/cincinnati-operator/blob/0731f08b2fbd3989b4642ff55c6a216e97d4c990/Makefile#L111

config/manifests/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+- ../default
+- ../samples
+- ../scorecard
```
wking, line 4: Where did `scorecard` come from?  I didn't have it [in my attempt at SDK v1.0.1][1].  And I don't see a `config/scorecard` in your tip:

```console
$ git ls-tree cde0dc77e7 -- config/
040000 tree ec4d134b363260017941b21a91707c809c363815    config/certmanager
040000 tree c5c218f132f09e55b2a1346c45a03961c3975373    config/crd
040000 tree 93494a3c9bdfd614b16c79e7cffe6bf5d4fb9ed6    config/default
040000 tree ebb074c993dfc68a5aa13c40221ae9d9b6c9aea6    config/manager
040000 tree 0b075d49b837e41817c0685134b3a7b3983c9232    config/manifests
040000 tree b535813978e2f2d3001990d684063a1d96bc9bed    config/olm-catalog
040000 tree ceedc8a2bee25b16deceb190b3ecfb02638ddfbc    config/prometheus
040000 tree 734fd7e3e4d2e67dc413916478fd062df5f21746    config/rbac
040000 tree ceece8f88259da3d69a81f5ffa8c82d5eb91d9c4    config/samples
040000 tree a878925b006459abe7238c43e5fe62c58d8174b2    config/webhook
```

[1]: https://github.com/openshift/cincinnati-operator/pull/79/files#diff-fe545782dbfdb2c00c396b25f7a8990e1d4af2d68d02e25363dc54c837f8aac6R3

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    alm-examples: |
```
wking, line 5: You should be able to [leave this blank][1] and have it inserted by `kustomize` magic at bundle-build time.

[1]: https://github.com/openshift/cincinnati-operator/pull/79/files#diff-d31bf97881e5efe5323e57c437e9999604ff7d312eef59930c415444b3e7c66aR5

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
...
+           "spec": {
+             "replicas": 1,
+             "releases": "openshift-release-dev/ocp-release",
+             "graphDataImage": "quay.io/cincinnati/demo-graph-data:for-testing-only"
+           }
+         }
+       ]
+    capabilities: Basic Install
+    description: Creates and maintains an OpenShift Container Platform Update Service instance
+    operators.operatorframework.io/builder: operator-sdk-v1.2.0
```
wking, line 22: I think we want [`operatorframework.io/suggested-namespace: openshift-update-service`][1] in here too.

[1]: https://github.com/openshift/cincinnati-operator/pull/79/files#diff-d31bf97881e5efe5323e57c437e9999604ff7d312eef59930c415444b3e7c66aR7

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
...
+           },
+           "spec": {
+             "replicas": 1,
+             "releases": "openshift-release-dev/ocp-release",
+             "graphDataImage": "quay.io/cincinnati/demo-graph-data:for-testing-only"
+           }
+         }
+       ]
+    capabilities: Basic Install
+    description: Creates and maintains an OpenShift Container Platform Update Service instance
```
wking, line 21: nit: `s/instance/operator./`?  Do we want to distinguish between the CSV installing the operator and subsequent UpdateService CRs as "instances"?

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
...
+    operators.operatorframework.io/builder: operator-sdk-v1.2.0
+    operators.operatorframework.io/project_layout: go.kubebuilder.io/v2
+  name: update-service-operator.v1.0.0
```
wking, line 24: I prefer [`openshift-update-service.vX.Y.Z`][1], to get closer to our 'OpenShift Update Service' official name.  No need to set an explicit version, some bundle-building magic will replace the `X.Y.Z` placeholder at build time.

[1]: https://github.com/openshift/cincinnati-operator/pull/79/files#diff-d31bf97881e5efe5323e57c437e9999604ff7d312eef59930c415444b3e7c66aR10

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
...
+spec:
+  apiservicedefinitions: {}
+  customresourcedefinitions:
+    owned:
+    - description: UpdateService is the Schema for the updateservices API.
+      displayName: Update Service
+      kind: UpdateService
+      name: updateservices.updateservice.operator.openshift.io
+      version: v1
+  description: |
```
wking, line 35: I dunno if a trailing newline would cause any problems, but everyone the existing CSV is including a [`-` block chomping indicator][1] (so `description: |-`):

```console
$ git grep -B2 'in a cluster is appealing' origin/pr/93 | grep description:
origin/pr/93:config/manifests/bases/update-service-operator.clusterserviceversion.yaml-  description: |
origin/pr/93:config/olm-catalog/updateservice-operator/1.0.0/updateservice-operator.v1.0.0.clusterserviceversion.yaml-  description: |-
```

[1]: https://yaml.org/spec/1.2/spec.html#id2794534

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
...
+    # Use Case
+    Running an Update Service instance in a cluster is appealing for offline OpenShift
+    clusters or for admins that want to provide their own graph data instead of
+    using [https://github.com/openshift/updateservice-graph-data](https://github.com/openshift/updateservice-graph-data/).
```
wking, line 39: s/updateservice-graph-data/cincinnati-graph-data/

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
...
+    Policy Engine in a chain will fetch its graph from the Graph Builder and the
+    last Policy Engine in a chain will serve the modified graph to the client.
+
+    An **Update Service client** is the end consumer of the release payloads. The
+    client periodically queries the Policy Engine for updates and applys them if
+    available.
+
+    # Query OpenShift's Update Service Endpoint
+    $ curl --silent --header 'Accept:application/json' 'https://api.openshift.com/api/upgrades_info/v1/graph?arch=amd64&channel=stable-4.2' | jq '. as $graph | $graph.nodes | map(.version == "4.2.13") | index(true) as $orig | $graph.edges | map(select(.[0] == $orig)[1]) | map($graph.nodes[.])'
+  displayName: Update Service
```
wking, line 65: `displayName: OpenShift Update Service`?

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
...
+  - Update Service
+  - UpdateService
+  - update-service
+  - Updates
+  - OSUS
+  - Cincinnati
+  links:
+  - name: Update Service Operator
+    url: https://github.com/openshift/cincinnati-operator
+  - name: OpenShift Container Platform Update Service
```
wking, line 92: Are we the OCP Update Service?  I thought we were just the `OpenShift Update Service`?

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
...
+    url: https://github.com/openshift/cincinnati
+  - name: OpenShift Graph Data
+    url: https://github.com/openshift/cincinnati-graph-data
+  maintainers:
+  - email: lmohanty@redhat.com
+    name: Lalatendu Mohanty
+  maturity: alpha
+  provider:
+    name: Red Hat
+  version: 1.0.0
```
wking, line 102: You can leave this at 0.0.0, and something substitutes the appropriate value at bundle-build time.

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
...
+    type: SingleNamespace
+  - supported: false
+    type: MultiNamespace
+  - supported: true
+    type: AllNamespaces
+  keywords:
+  - Update Service
+  - UpdateService
+  - update-service
+  - Updates
```
wking, line 86: s/Updates/Update/?  I dunno how the search engine behind these keywords handles pluralization, but I would expect singular keywords to be preferred and sufficient.

I think we also want to include `version` and `upgrade` in this list.  Not sure about keyword casing conventions either...

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,102 @@
...
+  links:
+  - name: Update Service Operator
+    url: https://github.com/openshift/cincinnati-operator
+  - name: OpenShift Container Platform Update Service
+    url: https://github.com/openshift/cincinnati
+  - name: OpenShift Graph Data
+    url: https://github.com/openshift/cincinnati-graph-data
+  maintainers:
+  - email: lmohanty@redhat.com
+    name: Lalatendu Mohanty
```
wking, line 98: Does this need to be a human?  Or can it be:

```yaml
  - email: aos-team-ota@redhat.com
    name: OpenShift Update Service maintainers
```

Makefile
```diff
@@ -35,3 +55,21 @@ unit-test:
...
+	cd config/manager && kustomize edit set image controller=$(IMG)
+	kustomize build config/manifests | operator-sdk generate bundle -q --overwrite --version $(BUNDLE_VERSION) $(BUNDLE_METADATA_OPTS)
+	operator-sdk bundle validate ./bundle
+
+# Build the bundle image.
+bundle-build:
+	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
```
LalatenduMohanty, line 75: I do not think we have defined the bundle.Dockerfile.

LalatenduMohanty: Also we need to build the bundled image in downstream as well. So somehow we need to just have one Dockerfile.

LalatenduMohanty (changes requested):

Makefile
jottofar: Channels are very much like our release candidate, fast, stable channels -> "Channels allow package authors to write different upgrade paths for different users (e.g. alpha vs. stable)"

I was trying to keep it simple for the first release and figured we would only have the one release initially.

Makefile
jottofar: It's defined as "semantic version of the operator in the generated bundle". How ever our operator version is a rather ugly git generated version. And as I recall this shows up in the hub as part of the meta data. I don't think we should use VERSION if the intent is to have it have different value for operator vs. the bundle.

Makefile
jottofar: I just simplified it and prefer explicitly doing installs but we can leave it as generated.

Makefile
jottofar: This is really TBD until what we know what we need to provide to the official build tools vs. what the official build tools do and produce. So much of what I have here is just the way I configured things to help me understand each step.

Makefile
jottofar: Again, I left some of that out until we know what we the developers run and provide through git to the build tools. The Makefile as generated seems to be very much oriented towards a developer running everything with no consideration or expectation of an official build. 

In general though, I preferred not having the dependency to kick off manifests and running it myself explicitly.

Makefile
LalatenduMohanty: I am not sure if the internal binary version shows up in the hub. Are you sure?

config/manifests/kustomization.yaml
jottofar: The `generate bundle` gets an error if it's not there. It gets created by v1.2 and I just copied it over to avoid the error.

Makefile
jottofar: I'm referring to the version that gets passed in below which is the bundle version which is defined as "semantic version of the operator in the generated bundle":

`operator-sdk generate bundle -q --overwrite --version $(BUNDLE_VERSION)`

And ends up here:
https://github.com/openshift/cincinnati-operator/blob/cde0dc77e7fadd5425e7be7e1b5c0a663e8bc423/config/manifests/bases/update-service-operator.clusterserviceversion.yaml#L102

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
jottofar: I believe this is describing what the operator installs. I simply changed `Cincinnati` to `OpenShift Container Platform Update Service`. We could change `instance` to `operand`?

Makefile
```diff
@@ -1,22 +1,42 @@
...
 
 # This is a placeholder for cincinnati-operator image placeholder
 # During development override this when you want to use an specific image
-# Example: IMG ?= quay.io/jottofar/updateservice-operator-index:v1
+# Example: IMG ?= quay.io/jottofar/update-service-operator:v1
 IMG ?= controller:latest
 
+BUNDLE_IMG ?= controller-bundle:latest
+
+CHANNELS ?= v1
```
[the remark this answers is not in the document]
jottofar: Channels are very much like our release candidate, fast, stable channels -> "Channels allow package authors to write different upgrade paths for different users (e.g. alpha vs. stable)"

Makefile
jottofar: `bundle.Dockerfile` gets generated by `operator-sdk generate bundle`, i.e. recipe `bundle`

Makefile
LalatenduMohanty: I see, that's makes sense. I think we should reuse `version` for bundle version. During build time we have to make sure use a meaningful version. Which is inline to what we have discussed till now.  Also we can change from commit id to something else in future as well.

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
wking: but the operator doesn't install any operands just landing the CSV.  You need to add a subsequent UpdateService CR to get any operand deployments.  I thought this property was supposed to describe the operator (e.g., maybe there are other operators who have no operand pods, they just manage a CRD and any subsequent CRs directly).  Unfortunately, the CSV declaration has [no docs for this property][1], so if we want to keep `description` about the operands for now, I'm fine with the text as it stands, or with `instance` changed to `operand(s)`.

[1]: https://github.com/operator-framework/api/blob/b55a341f6560db4adec39d69aab1ff3092ea202a/pkg/operators/v1alpha1/clusterserviceversion_types.go#L278

config/manifests/kustomization.yaml
wking: You're sure you don't have a `config/scorecard/` in your local checkout that just hasn't been committed?  I [had one][1].

[1]: https://github.com/openshift/cincinnati-operator/pull/79/files#diff-6aabd462c9a04e77fcde378dee12dda1b7988ac03c901e8eeb262bc03fbd6787R1

Makefile
wking: you can still do manual installs, with [this branch][1] using your already-installed command (regardless of the version it happens to be).  The scaffolded section just tries to install `controller-gen` if you don't have it in your path yet.

[1]: https://github.com/openshift/cincinnati-operator/blob/0731f08b2fbd3989b4642ff55c6a216e97d4c990/Makefile#L91

Makefile
wking: +1 to one `VERSION` and doing whatever we want to set it to a meaningful value at build time.  The commit ID is just a default for when folks don't care enough to set a meaningful version.

jottofar:
/retitle Operator bundle: changes to support operator bundling

Makefile
jottofar: done

Makefile
jottofar: Added installs back for controller-gen and kustomize as originally generated.

Makefile
jottofar: Added dependencies to `gen-for-bundle` on `manifest kustomize`

config/manifests/kustomization.yaml
jottofar: Right. I purposely removed it but it's required by bundling so have added back.

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
jottofar: done

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
jottofar: done

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
jottofar: I left it "as is" with only the operator name change. As originally written it is describing the operator.

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
jottofar: changed to `|-`

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
jottofar: changed

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
jottofar: done

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
jottofar: It's not case sensitive and filters on partial words so replaced `Updates` with `upgrade` and added `version`.

config/manifests/bases/update-service-operator.clusterserviceversion.yaml
jottofar: done

Makefile
jottofar: It can always be changed in the tmp area before it's copied to distgit.

Makefile
```diff
@@ -35,3 +52,53 @@ unit-test:
 
 build: $(SOURCES)
 	go build $(GOBUILDFLAGS) -ldflags="$(GOLDFLAGS)" -o ./update-service-operator ./
+
+# find or download controller-gen
+# download controller-gen if necessary
+controller-gen:
+ifeq (, $(shell which controller-gen))
+        @{ \
```
wking, line 60: [images][1]:

```
Makefile:60: *** missing separator (did you mean TAB instead of 8 spaces?).  Stop.
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-operator/93/pull-ci-openshift-cincinnati-operator-master-images/1334930754748026880#1:build-log.txt%3A71

jottofar: I did. Bit by cut-and-paste once again :) Fixed in [27ac93d](https://github.com/openshift/cincinnati-operator/pull/93/commits/27ac93d60d45be1194a27a32f9e525801d2456d3)

wking:
CI is green.  If we need any follow-up fixes (I can't think of any at the moment), we can land them in later PRs.

/lgtm

PR #95, opened 2020-12-09
Adding SDK version to the operator

LalatenduMohanty:
As for creating the operator bundle we need to know which version of
 SDK should be used to generate the manifests

### DIFF
--- a/main.go
+++ b/main.go
@@ -47,9 +47,10 @@ func init() {
 }
 
 func printVersion() {
-	log.Info(fmt.Sprintf("Operator Version: %s", version.Operator))
-	log.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
-	log.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
+	log.Info(fmt.Sprintf("version: v%s", version.VersionOperator))
+	log.Info(fmt.Sprintf("operator-sdk version: v%s", version.VersionSDK))
+	log.Info(fmt.Sprintf("go Version: %s", runtime.Version()))
+	log.Info(fmt.Sprintf("go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
 }
 
 // getWatchNamespace returns the Namespace the operator should be watching for changes
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 # Current Operator version
 VERSION ?= 0.0.1
+VERSIONSDK ?= 1.2.0
 
 .PHONY: \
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -47,9 +47,10 @@ func init() {
 }
 
 func printVersion() {
-	log.Info(fmt.Sprintf("Operator Version: %s", version.Operator))
-	log.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
-	log.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
+	log.Info(fmt.Sprintf("version: v%s", version.VersionOperator))
```
jottofar, line 50: Why the change from `Operator Version`?

LalatenduMohanty: You mean from `version.Operator` to `version.versionOperator` ? to keep it consistent with `Version.OperatorSDK`.

jottofar: Why did you change the label string from "Operator Version:" which seems more clear?

LalatenduMohanty: Will change ti to `operator version`

wking: I am in favor of the `Makefile` change, but would rather keep this out of Go, see [here][1] and that commit message.  Is there a benefit to storing this information in Go?

[1]: https://github.com/openshift/cincinnati-operator/commit/c95c673082138ce8a6efb92b789791a952e4dde4#diff-f6bf0ae0d21c9408c6624174da11dafb369ce306f2cdcea704285d8acc87e19eL5

main.go
```diff
@@ -47,9 +47,10 @@ func init() {
...
+	log.Info(fmt.Sprintf("operator-sdk version: v%s", version.VersionSDK))
```
wking, line 51: I am in favor of the `Makefile` change, but would rather keep this out of Go, see [here][1] and that commit message.  Is there a benefit to storing this information in Go?

[1]: https://github.com/openshift/cincinnati-operator/commit/c95c673082138ce8a6efb92b789791a952e4dde4#diff-f6bf0ae0d21c9408c6624174da11dafb369ce306f2cdcea704285d8acc87e19eL5

main.go
```diff
@@ -47,9 +47,10 @@ func init() {
...
+	log.Info(fmt.Sprintf("go Version: %s", runtime.Version()))
+	log.Info(fmt.Sprintf("go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
```
wking, line 53: Why the downcasing?  The Go project name is canonically title case, e.g. see examples [here][1].

[1]: https://golang.org/doc/

PR #96, opened 2021-01-19
document using shorter name for Update Service CR

PratikMahajan:
In case the user creates the Cincinnati CR with a longer name, there
might be a case where the route's hostname exceeds 63 chars. This will
result in the route not getting created, inturn crashing the operator.

Document using a shorter name for Cincinnati CR to prevent this
issue.

### DIFF
--- a/docs/disconnected-cincinnati-operator.md
+++ b/docs/disconnected-cincinnati-operator.md
@@ -240,6 +240,9 @@ You might want to review the documentation around disconnected registries to lea
 
     gives a full URI.
 
+    If you see `ReconcileCompleted` status as `false` with reason `CreateRouteFailed` caused by `host must conform to DNS 1123 naming convention`
+    and `must be no more than 63 characters`, try creating the Update Service with a shorter name.
+    
 4. At this point we have Cincinnati working locally, but our cluster is still pointing to the public Cincinnati instance as we can see in the image below
 
     ![Cincinnati Public URL](assets/public-cincinnati.png)


docs/disconnected-cincinnati-operator.md
```diff
@@ -240,6 +240,8 @@ You might want to review the documentation around disconnected registries to lea
 
     gives a full URI.
 
+    If you see `host name validation error` in your UpdateService's status, try creating an UpdateService with a shorter name.
```
LalatenduMohanty, line 243: s/UpdateService/OpenShift update service/. Also we should exactly mention the limit i.e. 63 characters.

LalatenduMohanty (changes requested):

docs/disconnected-cincinnati-operator.md
```diff
@@ -240,6 +240,9 @@ You might want to review the documentation around disconnected registries to lea
 
     gives a full URI.
 
+    If you see `ReconcileCompleted` status as `false` with reason `CreateRouteFailed` caused by `host must conform to DNS 1123 naming convention`
+    and `must be no more than 63 characters`, try creating the Update Service with a shorter name. The resource name should be 63 characters or less.
```
wking, line 244: > The resource name should be 63 characters or less.

"The resource" is a bit ambiguous (UpdateService?  Route?).  And really, it's the combination of the Route's name and namespace.  Can we rely on the quoted `must be no more than 63 characters` to address @LalatenduMohanty's [request for the 63-char mention][1], rely on the rest of the Route-tooling error message for further context, and drop the "The resource..." sentence?

Also nit:

```console
$ git --no-pager show --oneline --check origin/pr/96
070a7a0 (origin/pr/96) document using shorter name for cincinnati cr
docs/disconnected-cincinnati-operator.md:244: trailing whitespace.
+    and `must be no more than 63 characters`, try creating the Update Service with a shorter name. The resource name should be 63 characters or less. 
docs/disconnected-cincinnati-operator.md:245: trailing whitespace.
+    
```

[1]: https://github.com/openshift/cincinnati-operator/pull/96#discussion_r562803709

LalatenduMohanty: I do not see any reason to drop the sentence because this is in the documentation and there is no harm if we add little more information.

wking: > I do not see any reason to drop the sentence...

I don't think it adds anything to the `host must conform to DNS 1123 naming convention` and `must be no more than 63 characters` quoted in the previous sentence.  But whatever, if we keep it, can we at least clarify "The resource" to "The UpdateService" or whatever it was intended to be?

PratikMahajan: Removed "The resource" line, as it would have caused more confusion than it was solving.

wking (approved):
/lgtm

PR #99, opened 2021-02-22
updates OpenShift doc links to current version 4.6

mhrivnak:


wking:
I dunno why Prow doesn't like us.  Manually added the approved label :p.

PR #103, opened 2021-03-19
Bug 1939788: config/samples: Shorten name

jottofar:
To prevent invalid host error upon route creation. Also replaced incorrect generated spec contents with defaulted "replicas" value.


LalatenduMohanty (approved):
/lgtm

PR #116, opened 2021-04-19
controllers/updateservice_controller: Handle GET/CREATE races

wking:
Handle [races][1] [like][2]:

```
2021-04-13T17:29:01.654Z	ERROR	controller_updateservice	CreateRouteFailed	{"Request.Namespace": "openshift-updateservice", "Request.Name": "example", "error": "routes.route.openshift.io \"example-policy-engine-route\" already exists"}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/src/github.com/openshift/cincinnati-operator/vendor/github.com/go-logr/zapr/zapr.go:132
github.com/openshift/cincinnati-operator/controllers.handleErr
	/go/src/github.com/openshift/cincinnati-operator/controllers/updateservice_controller.go:165
github.com/openshift/cincinnati-operator/controllers.(*UpdateServiceReconciler).ensurePolicyEngineRoute
	/go/src/github.com/openshift/cincinnati-operator/controllers/updateservice_controller.go:481
github.com/openshift/cincinnati-operator/controllers.(*UpdateServiceReconciler).Reconcile
```

where we are racing ourselves to create the route:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati-operator/112/pull-ci-openshift-cincinnati-operator-release-4.6-operator-e2e/1381991628536811520/artifacts/operator-e2e/gather-audit-logs/artifacts/audit-logs.tar | tar -xz
$ cd registry-*
$ zgrep -h '"resource":"routes".*"name":".*example.*"' audit_logs/*/*.gz 2>/dev/null | jq -r '.stageTimestamp + " " + .verb + " " + .kind + " " + (.responseStatus | tostring) + " " + .user.username' | sort
2021-04-13T17:29:01.619640Z create Event {"metadata":{},"code":201} system:serviceaccount:openshift-updateservice:updateservice-operator
2021-04-13T17:29:01.646645Z create Event {"metadata":{},"status":"Failure","reason":"AlreadyExists","code":409} system:serviceaccount:openshift-updateservice:updateservice-operator
2021-04-13T17:29:01.873911Z update Event {"metadata":{},"code":200} system:serviceaccount:openshift-ingress:router
2021-04-13T17:29:01.874271Z update Event {"metadata":{},"code":200} system:serviceaccount:openshift-ingress:router
2021-04-13T17:29:01.962248Z update Event {"metadata":{},"status":"Failure","reason":"Conflict","code":409} system:serviceaccount:openshift-ingress:router
2021-04-13T17:29:01.962785Z update Event {"metadata":{},"code":409} system:serviceaccount:openshift-ingress:router
2021-04-13T17:29:02.817842Z update Event {"metadata":{},"code":200} system:serviceaccount:openshift-updateservice:updateservice-operator
2021-04-13T17:29:02.844601Z update Event {"metadata":{},"code":200} system:serviceaccount:openshift-updateservice:updateservice-operator
...
```

With this commit, the AlreadyExists 409 will go from a `CreateRouteFailed` error to a recursive attempt at reconciliation.  I sleep for a second before the recursive attempt to give the caching layer time to catch up, so the expectation is that by the first or second recursion, the initial GET will successfuly return the in-cluster resource, and we can procede to compare with our desired content and UPDATE if necessary.

I then went through and applied that same pattern to the rest of our resources.  As part of the unification, I centralized `SetControllerReference` and `handleErr` calls in the lower level.  For example, the various `ensureConfigMap` callers now share central reference and error handling, instead of repeating that code locally.

The pattern-unification also caught some unhandled errors, like the PDB and Route UPDATE calls and the Route `IngressURI` call.  With this commit, failures for those calls will abort subsequent resource reconciliation.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-operator/112/pull-ci-openshift-cincinnati-operator-release-4.6-operator-e2e/1381991628536811520
[2]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati-operator/112/pull-ci-openshift-cincinnati-operator-release-4.6-operator-e2e/1381991628536811520/artifacts/operator-e2e/gather-extra/artifacts/pods/openshift-updateservice_updateservice-operator-7675564bcd-r8vzc_updateservice-operator.log

### DIFF
--- a/controllers/updateservice_controller.go
+++ b/controllers/updateservice_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"time"
 
 	"github.com/go-logr/logr"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
@@ -29,6 +30,11 @@ import (
 )
 
 var log = logf.Log.WithName("controller_updateservice")
+
+// createConflictSleep is the duration to sleep before trying again
+// after a GET fails with NotFound but a subsequent CREATE fails with
+// AlreadyExists.
+var createConflictSleep = time.Second
 
 // blank assignment to verify that ReconcileUpdateService implements reconcile.Reconciler
 var _ reconcile.Reconciler = &UpdateServiceReconciler{}
@@ -235,17 +241,7 @@ func (r *UpdateServiceReconciler) ensurePullSecret(ctx context.Context, reqLogge
 		return nil
 	}
 
-	// Set UpdateService instance as the owner and controller
-	if err := controllerutil.SetControllerReference(instance, resources.pullSecret, r.Scheme); err != nil {
-		return err
-	}
-
-	if err := r.ensureSecret(ctx, reqLogger, resources.pullSecret); err != nil {
-		handleErr(reqLogger, &instance.Status, "EnsureSecretFailed", err)
-		return err
-	}
-
-	return nil
+	return r.ensureSecret(ctx, reqLogger, instance, resources.pullSecret)
 }
 
 func (r *UpdateServiceReconciler) ensureAdditionalTrustedCA(ctx context.Context, reqLogger logr.Logger, instance *cv1.UpdateService, resources *kubeResources) error {
@@ -267,11 +263,7 @@ func (r *UpdateServiceReconciler) ensureAdditionalTrustedCA(ctx context.Context,
 		return err
 	}
 
-	if err := r.ensureConfigMap(ctx, reqLogger, resources.trustedCAConfig); err != nil {
-		handleErr(reqLogger, &instance.Status, "EnsureConfigMapFailed", err)
-		return err
-	}
-	return nil
+	return r.ensureConfigMap(ctx, reqLogger, instance, resources.trustedCAConfig)
 }
 
 func (r *UpdateServiceReconciler) ensureDeployment(ctx context.Context, reqLogger logr.Logger, instance *cv1.UpdateService, resources *kubeResources) error {
@@ -285,11 +277,15 @@ func (r *UpdateServiceReconciler) ensureDeployment(ctx context.Context, reqLogge
 
 	if err != nil && errors.IsNotFound(err) {
 		reqLogger.Info("Creating Deployment", "Namespace", deployment.Namespace, "Name", deployment.Name)
-		err := r.Client.Create(ctx, deployment)
-		if err != nil {
+		if err = r.Client.Create(ctx, deployment); err != nil && errors.IsAlreadyExists(err) {
+			reqLogger.Info("Deployment already exists, reconciling again", "Namespace", deployment.Namespace, "Name", deployment.Name)
+			time.Sleep(createConflictSleep)
+			return r.ensureDeployment(ctx, reqLogger, instance, resources)
+		} else if err != nil {
 			handleErr(reqLogger, &instance.Status, "CreateDeploymentFailed", err)
+			return err
 		}
-		return err
+		return nil
 	} else if err != nil {
 		handleErr(reqLogger, &instance.Status, "GetDeploymentFailed", err)
 		return err
@@ -364,8 +360,7 @@ func (r *UpdateServiceReconciler) ensureDeployment(ctx context.Context, reqLogge
 
 	if !reflect.DeepEqual(updated.Spec, found.Spec) {
 		reqLogger.Info("Updating Deployment", "Namespace", deployment.Namespace, "Name", deployment.Name)
-		err = r.Client.Update(ctx, updated)
-		if err != nil {
+		if err = r.Client.Update(ctx, updated); err != nil {
 			handleErr(reqLogger, &instance.Status, "UpdateDeploymentFailed", err)
 			return err
 		}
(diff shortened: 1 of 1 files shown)


mhrivnak:
Something about this bug doesn't seem right. The controller-runtime client is heavily used in similar patterns, and sleep statements should not be necessary.

It looks like the Route is the last resource that is "ensured". From the log file it looks like during the second Reconcile, all the other resources were correctly identified as existing, and the Route is the only one that it tried to create a second time. I don't see anything obvious to explain that on quick glance through the code, but there might be some explanation. It seems a real shame to add a bunch of sleep statements if there's a possibility to find the root cause.

Maaaaaybe if the Route is implemented with an aggregated API server, that could be slower to catch up than the rest of the resources? In that case it might make sense to just ignore an "already exists" error and have the knowledge that whenever the Route does actually have a recognized change, an event will fire and cause another Reconciliation at that time, which will be reflected in the cache.

Could be interesting to ask in the operators slack channel if anyone's seen something similar.

wking:
```console
$ zgrep -h '"resource":"routes".*"name":".*example.*"' audit_logs/*/*.gz 2>/dev/null | jq -r '.requestReceivedTimestamp + " " + .stageTimestamp + " " + (.sourceIPs | tostring) + " " + .verb + " " + (.responseStatus | tostring)' | sort | head -n2
2021-04-13T17:29:01.364379Z 2021-04-13T17:29:01.619640Z ["10.0.32.3","10.128.0.1"] create {"metadata":{},"code":201}
2021-04-13T17:29:01.638530Z 2021-04-13T17:29:01.646645Z ["10.0.32.3","10.128.0.1"] create {"metadata":{},"status":"Failure","reason":"AlreadyExists","code":409}
```

shows that even though there are only 50ms between `stageTimestamp` for the two calls, the second call was still launched with a `requestReceivedTimestamp` later than the first's `stageTimestamp`.

> ... sleep statements should not be necessary.

I agree.  I think the issue here is probably a client-side cache miss on the second GET, because we shouldn't even be hitting the API-server for a GET that's milliseconds after a successful CREATE.  But I don't have time to go rooting around in the caching logic to figure out what's going wrong there.  The one-second sleep should only come into effect when we have that cache bug, and we'll have logs that we can look at to ensure we aren't hitting the sleep more than we expect.  We can drop it once someone bottoms out the caching issue.

> Maaaaaybe if the Route is implemented with an aggregated API server...

Routes are OpenShift-specific, and I think they are implemented via an aggregated OpenShift API-server.  Not my space though, so don't take me too seriously.

> In that case it might make sense to just ignore an "already exists"...

I don't like `nil` responses unless the reconciler is happy with its view of the current state.  The recursive call is intended to retrieve the current in-cluster state and compare it to the desired state.  The current noisy stack traces are followed very shortly by successful (stuttering?) updates:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati-operator/112/pull-ci-openshift-cincinnati-operator-release-4.6-operator-e2e/1381991628536811520/artifacts/operator-e2e/gather-extra/artifacts/pods/openshift-updateservice_updateservice-operator-7675564bcd-r8vzc_updateservice-operator.log | grep -i route
2021-04-13T17:29:01.357Z        INFO    controller_updateservice        Creating Route  {"Request.Namespace": "openshift-updateservice", "Request.Name": "example", "Namespace": "openshift-updateservice", "Name": "example-policy-engine-route"}
2021-04-13T17:29:01.636Z        INFO    controller_updateservice        Creating Route  {"Request.Namespace": "openshift-updateservice", "Request.Name": "example", "Namespace": "openshift-updateservice", "Name": "example-policy-engine-route"}
2021-04-13T17:29:01.654Z        ERROR   controller_updateservice        CreateRouteFailed       {"Request.Namespace": "openshift-updateservice", "Request.Name": "example", "error": "routes.route.openshift.io \"example-policy-engine-route\" already exists"}
github.com/openshift/cincinnati-operator/controllers.(*UpdateServiceReconciler).ensurePolicyEngineRoute
2021-04-13T17:29:01.662Z        ERROR   controller      Reconciler error        {"reconcilerGroup": "updateservice.operator.openshift.io", "reconcilerKind": "UpdateService", "controller": "updateservice", "name": "example", "namespace": "openshift-updateservice", "error": "routes.route.openshift.io \"example-policy-engine-route\" already exists"}
2021-04-13T17:29:02.663Z        INFO    controller_updateservice        Updating Route  {"Request.Namespace": "openshift-updateservice", "Request.Name": "example", "Namespace": "openshift-updateservice", "Name": "example-policy-engine-route"}
2021-04-13T17:29:02.834Z        INFO    controller_updateservice        Updating Route  {"Request.Namespace": "openshift-updateservice", "Request.Name": "example", "Namespace": "openshift-updateservice", "Name": "example-policy-engine-route"}
2021-04-13T17:36:43.647Z        INFO    controller_updateservice        Updating Route  {"Request.Namespace": "openshift-updateservice", "Request.Name": "example", "Namespace": "openshift-updateservice", "Name": "example-policy-engine-route"}
2021-04-13T17:36:43.737Z        INFO    controller_updateservice        Updating Route  {"Request.Namespace": "openshift-updateservice", "Request.Name": "example", "Namespace": "openshift-updateservice", "Name": "example-policy-engine-route"}
2021-04-13T17:36:43.808Z        INFO    controller_updateservice        Updating Route  {"Request.Namespace": "openshift-updateservice", "Request.Name": "example", "Namespace": "openshift-updateservice", "Name": "example-policy-engine-route"}
```

That's distracting.  This PR will reduce the failed-CREATE distraction while still restricting `nil` to "we just looked at or touched the in-cluster state, and it's fine".  We're not in a huge hurry, so it's not a problem to sleep a second to avoid some kind of weird request flood if for some reason the API-server is completely broken about this.  If we already have some reasonable client-side throttling in place, I'll happily drop the sleeps.

mhrivnak:
> If we already have some reasonable client-side throttling in place, I'll happily drop the sleeps.

There is rate limiting with backoff if you return `Requeue: true`. Otherwise you could just return and wait for the event that corresponds to the Route's creation to get to the controller, cause an update to the cache, and run another Reconcile.

PR #118, opened 2021-04-21
Increasing the go test timeout to 20 minutes.

LalatenduMohanty:
The CI kept failing in release-4.6 branch and the root cause is the
sracpping time graph-builder container in OSUS operand which is around
 8 minutes or so. However this adds the
total test duration more than 10 mins which causing the test to fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ deploy:
 func-test: deploy
 	@echo "Running functional test suite"
 	go clean -testcache
-	go test -v ./functests/...
+	go test -timeout 20m -v ./functests/...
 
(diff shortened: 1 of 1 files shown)


wking (approved):
/lgtm

LalatenduMohanty:
Also this bump is harmless as we have another timeout [1] for specific tests which we are not changing. 
```
timeout            = time.Second * 600
```
[1] https://github.com/openshift/cincinnati-operator/blob/master/functests/utils.go#L30

wking (approved):
/lgtm

PR #119, opened 2021-04-21
Expire all test results in the go build cache before new run

LalatenduMohanty:
Else go test was giving success everytime even if the test cluster was
deleted as the result was cached

$ go test -v ./functests/...
=== RUN   TestCustomResource
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
--- PASS: TestCustomResource (241.87s)
PASS
ok  	github.com/openshift/cincinnati-operator/functests	(cached)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,7 @@ endif
 
 clean:
 	@echo "Cleaning previous outputs"
+	go clean -testcache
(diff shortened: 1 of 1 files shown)


vrutkovs (approved):
/lgtm

PR #134, opened 2021-10-14
OTA-378: Refine OSUS Operator for better UX and error proofing

jottofar:
Changes:

1. Operator will only `Reconcile` operands (OSUS apps) created, or already running, in the same namespace as the operator.
2. Disabled AllNamespaces installation mode since this PR restricts OSUS operator namespace watching to `openshift-config` and the namespace in which the operator is installed.

### DIFF
--- a/config/manifests/bases/update-service-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/update-service-operator.clusterserviceversion.yaml
@@ -66,11 +66,11 @@ spec:
   installModes:
   - supported: true
     type: OwnNamespace
-  - supported: true
+  - supported: false
     type: SingleNamespace
   - supported: false
     type: MultiNamespace
-  - supported: true
+  - supported: false
     type: AllNamespaces
   keywords:
   - Update Service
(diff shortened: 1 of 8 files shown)


config/manifests/bases/update-service-operator.clusterserviceversion.yaml
```diff
@@ -66,11 +66,11 @@ spec:
   installModes:
   - supported: true
     type: OwnNamespace
-  - supported: true
+  - supported: false
     type: SingleNamespace
   - supported: false
     type: MultiNamespace
-  - supported: true
+  - supported: false
```
wking, line 73: What does this do for folks who are updating from a previous release that they'd installed in `AllNamespaces` mode?

jottofar: The operator doesn't support seamless upgrades so it may be a matter of uninstalling the current operator and then installing the later version at which time it will be bounded by its restrictions. When AllNamespaces install mode is selected installation is limited to `openshift-operators` and `openshift-update-service`. When AllNamespaces install mode is not supported the operator will not install successfully into `openshift-operators`. I created [OTA-506](https://issues.redhat.com/browse/OTA-506) to document such things.

PR #137, opened 2021-12-10
Update disconnected UpdateService operator documentation

karampok:
Hi,  I was doing a cluster update using the cincinnati-operator in a disconnected cluster. I found that the doc is a bit outdated and therefore here is an updated version.

PR is split into two commits, separating the rename file  in case this makes it easier for reviewing versus the older version.
(I will squash before merging)

Thx!


docs/disconnected-updateservice-operator.md
```diff
@@ -0,0 +1,295 @@
...
+---
+apiVersion: updateservice.operator.openshift.io/v1
+kind: UpdateService
+metadata:
+  name: karampok
+  namespace: openshift-update-service
+spec:
+  replicas: 1
+  releases: "{DISCONNECTED_REGISTRY}/ocp4/release"
+  #releases: "{DISCONNECTED_REGISTRY}/ocp4 // OCP Release Image and Release Content in the same path
```
vrutkovs, line 158: Lets remove this, we should never recommend mixing release images and content in one repo, as it would most certainly make image scraping complicated

docs/disconnected-updateservice-operator.md
```diff
@@ -0,0 +1,295 @@
...
+
+```
+skopeo copy docker://${DISCONNECTED_REGISTRY}/ocp4:4.5.2-x86_64 docker://${DISCONNECTED_REGISTRY}/release:4.5.2-x86_64 --authfile=/path/to/pull_secret.json
+skopeo copy docker://${DISCONNECTED_REGISTRY}/ocp5:4.5.3-x86_64 docker://${DISCONNECTED_REGISTRY}/release:4.5.3-x86_64 --authfile=/path/to/pull_secret.json
+```
+
+#### CreateRouteFailed
+
+    
+If you see `ReconcileCompleted` status as `false` with reason `CreateRouteFailed` caused by `host must conform to DNS 1123 naming convention`
```
vrutkovs, line 180: Is it worth mentioning? It looks like a fairly common error

docs/disconnected-updateservice-operator.md
```diff
@@ -0,0 +1,295 @@
...
+and `must be no more than 63 characters`, try creating the Update Service with a shorter name.
+    
+## Make client cluster use UpdateService instance
+
+Once UpdateService starts, there is a public URL (OCP route) available. This URL is should be configured in the client cluster.
```
vrutkovs, line 185: Needs a link to "update ClusterVersion's url" doc section

docs/disconnected-updateservice-operator.md
```diff
@@ -0,0 +1,295 @@
...
+
+POLICY_ENGINE_GRAPH_URI="$(oc -n openshift-update-service get updateservice karampok -o jsonpath='{.status.policyEngineURI}/api/upgrades_info/v1/graph')"
+PATCH="{\"spec\":{\"upstream\":\"${POLICY_ENGINE_GRAPH_URI}\"}}"
+oc patch clusterversion version -p $PATCH --type merge
+```
+
+
+### Make client cluster trust UpdateService server certificate
+
+Unless your UpdateService tls server certificate is somehow trusted by the client OCP cluster (e.g. well known root CA), then you probably get the
```
vrutkovs, line 207: Needs a link to addtionalTrustedCA doc section from proxy server

docs/disconnected-updateservice-operator.md
```diff
@@ -0,0 +1,295 @@
...
+    oc create configmap trusted-updateservice-ca -n openshift-config --from-file=ca-bundle.crt=OCPRootCA.pem
+    oc patch proxy/cluster -p '{"spec":{"trustedCA":{"name":"trusted-updateservice-ca"}}}' --type merge
+    oc edit proxy/cluster
+    # ... wait until pods are restarting ..
+```
+
+Finally you should get the following:
+
+```
+dan adm upgrade
```
vrutkovs, line 235:
```suggestion
oc adm upgrade
```

docs/disconnected-updateservice-operator.md
```diff
@@ -0,0 +1,295 @@
...
+>  Reason: AdminAckRequired
+>  Message: Kubernetes 1.22 and therefore OpenShift 4.9 remove several APIs which require admin consideration. Please see the knowledge article https://access.redhat.com/articles/6329921 for details and instructions.
+> Updates:
+> VERSION IMAGE
+> 4.8.23  {DISCONNECTED_REGISTRY}/ocp4/release@sha256:3fab205d36c66825423274eac90f4c142a18cdf358b4a666a1783d325afba860
+```
+
+## Update Client Cluster
+
+We can force the update using the CLI
```
vrutkovs, line 247: Lets not mention `force` as the cluster may lose supported status. Needs to be reformatted into `in case of emergency` or smth similar. 
Same for "--allow-upgrade-with-warnings".

"--to-latest" is also probably not a great idea, lets rework into "to desired version" and adding that "--to-latest" is a shortcut for that

docs/disconnected-updateservice-operator.md
```diff
@@ -0,0 +1,295 @@
...
+
+```
+oc adm upgrade --to-latest --allow-upgrade-with-warnings --force
+```
+    
+### Add release signature
```
vrutkovs, line 253:
```suggestion
### Add release signature verification config
```

docs/disconnected-updateservice-operator.md
```diff
@@ -0,0 +1,295 @@
...
+
+To make the release safe, so no extra flags are needed, we can add a release signature configmap:
```
vrutkovs, line 255: This needs to be moved before "start the upgrade". Signature verification is important to prevent supply chain attacks

docs/external-registry-ca.md
```diff
@@ -17,7 +17,7 @@ kind: ConfigMap
 metadata:
   name: trusted-ca
 data:
-  cincinnati-registry: |
+  updateservice-registry: |
```
vrutkovs, line 20: Not sure if its needed. IIUC this key depends on the CR name, @wking?

vrutkovs (changes requested):

docs/disconnected-updateservice-operator.md
mvazquezc: This depends on the base domain of the cluster, it's a limitation on the route names length. I believe keeping it here doesn't hurt, but up to you.

docs/external-registry-ca.md
karampok: From observation it is needed, https://github.com/openshift/cincinnati-operator/blob/a3107f885a3290f10ae24147c27930bbc911f1a0/controllers/names.go#L19-L20

docs/disconnected-updateservice-operator.md
vrutkovs: Sounds like its a common problem, so lets keep it - it needs a link to bugzilla though

docs/disconnected-updateservice-operator.md
```diff
@@ -0,0 +1,296 @@
...
+
+## Update Client Cluster
+
+We can start the update using the CLI and executing:
+
+```
+oc adm upgrade --to=<version>
+```
+    
+In case of emergency where the update is being blocked, the argument `--allow-upgrade-with-warnings` can be used.
```
vrutkovs, line 273:
```suggestion
If this command throws any warnings which are not important for cluster upgrade, use the argument `--allow-upgrade-with-warnings`. If user wants to upgrade the cluster to a cluster which is not available in the graph, use `--force` argument. Note, that cluster upgraded using any of these arguments may be in unsupportable state
```

vrutkovs (commented):
Looks good, lets run it by @wking to confirm I didn't miss anything

docs/disconnected-updateservice-operator.md
vrutkovs: If operator allows them to configure settings which end up in an invalid route I'd say its a bug

karampok: I do not think there is bug here, if I understand it is just a small reminder that 
basedomain + object name is the DNS name which should feed the length.

karampok:
Hi, I cannot see a reason why the CI job is failing because of my changes, let me know if there is anything else missing from my side. Thx

vrutkovs:

Doc changes only

karampok:
Hi! Any feedback on this one? Thank you

vrutkovs:
LGTM, lets squash it in one commits so that I could merge this?

karampok:
Squashed to one commit, thanks!

vrutkovs (approved):
/lgtm

PR #139, opened 2022-01-05
Use http probe for policy-engine

vrutkovs:
This allows us to use keep-alived connection to check that policy-engine is still running. This becomes important under load, when a new liveness probe connection becomes expensive.

### DIFF
--- a/controllers/new.go
+++ b/controllers/new.go
@@ -594,8 +594,10 @@ func (k *kubeResources) newPolicyEngineContainer(instance *cv1.UpdateService, im
 			PeriodSeconds:       10,
 			TimeoutSeconds:      3,
 			Handler: corev1.Handler{
-				TCPSocket: &corev1.TCPSocketAction{
-					Port: intstr.FromInt(8081),
+				HTTPGet: &corev1.HTTPGetAction{
+					Path:   "/metrics",
+					Port:   intstr.FromInt(9081),
+					Scheme: corev1.URISchemeHTTP,
 				},
 			},
 		},


controllers/new.go
```diff
@@ -594,8 +594,10 @@ func (k *kubeResources) newPolicyEngineContainer(instance *cv1.UpdateService, im
 			PeriodSeconds:       10,
 			TimeoutSeconds:      3,
 			Handler: corev1.Handler{
-				TCPSocket: &corev1.TCPSocketAction{
-					Port: intstr.FromInt(8081),
```
wking, line 598: I was looking into the backstory here, and we've been using TCP sockets [to avoid issues with missing query parameters][1].  That would still be a problem today:

```
$ curl -i https://api.openshift.com/api/upgrades_info/graph
HTTP/2 400 
content-length: 80
content-type: application/json
date: Wed, 05 Jan 2022 18:49:07 GMT
set-cookie: 609f91be912da47026eb733a4b0de2df=a9edca1136b8c149892e9ea5dae71a33; path=/; HttpOnly; Secure; SameSite=None
x-envoy-upstream-service-time: 18
server: envoy

{"kind":"missing_params","value":"mandatory client parameters missing: channel"}
```

But since then, openshift/cincinnati#181 made `version` optional.  And it looks like `arch` and `id` have always been optional, and Cincinnati doesn't care if we pick a non-existent channel:

```console
$ curl 'https://api.openshift.com/api/upgrades_info/graph?channel=does-not-exist'
{"version":1,"nodes":[],"edges":[],"conditionalEdges":[]}
```

So it seems like we could use:

```go
HTTPGet: &corev1.HTTPGetAction{
  Path:   "api/upgrades_info/graph?channel=does-not-exist",
  Port:   intstr.FromInt(8081),
  Scheme: corev1.URISchemeHTTP,
  // optional HTTPHeaders setting Accept: application/vnd.redhat.cincinnati.v1+json
}
```

if we wanted to avoid `/metrics` and stay closer to the client pathways serving actual graphs.

[1]: https://github.com/openshift/cincinnati/pull/83/commits/db9ad4a1c7434fbc82088eb7efd61c40953e1ff3

controllers/new.go
```diff
@@ -594,8 +594,10 @@ func (k *kubeResources) newPolicyEngineContainer(instance *cv1.UpdateService, im
...
+				HTTPGet: &corev1.HTTPGetAction{
+					Path:   "/metrics",
+					Port:   intstr.FromInt(9081),
```
wking, line 599: [e2e][1]:

```
I0105 17:21:47.971812    5751 utils.go:106] Waiting for full availability of example deployment (0/1)
    updateservice_creation_test.go:81: timed out waiting for the condition
```

Because these probes are 503ing:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati-operator/139/pull-ci-openshift-cincinnati-operator-master-operator-e2e/1478769955875852288/artifacts/operator-e2e/gather-extra/artifacts/nodes/ci-op-in4lfpdq-fb094-vx4mk-worker-c-wwvk4/journal | zgrep 'example.*probe' | tail -n1
Jan 05 17:21:48.150903 ci-op-in4lfpdq-fb094-vx4mk-worker-c-wwvk4 hyperkube[1425]: I0105 17:21:48.150428    1425 prober.go:116] "Probe failed" probeType="Readiness" pod="openshift-updateservice/example-56cc96c9fc-ktflx" podUID=7ccae62d-aec1-4c95-99bf-1d078ac1e248 containerName="graph-builder" probeResult=failure output="HTTP probe failed with statuscode: 503"
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-operator/139/pull-ci-openshift-cincinnati-operator-master-operator-e2e/1478769955875852288#1:build-log.txt%3A80

vrutkovs: >containerName="graph-builder"

That's GB probe, not PE

vrutkovs: GB initial scrape gets stuck at
>[2022-01-06T10:12:45Z ERROR graph_builder::graph] Concrete release with the same version (4.7.32+amd64) already exists

vrutkovs: Checking if updated Cincinnati would resolve that - https://github.com/openshift/release/pull/25027

wking (approved):
/lgtm

PR #140, opened 2022-01-10
test: delete CR on successful pass only

vrutkovs:
This would make gather-* steps collect the state of deployment in prow job artifacts when operator-e2e fails


wking (approved):
/lgtm

PR #141, opened 2022-01-11
deploy.sh: show image pull specs when deploying

vrutkovs:
Removed openshift CI case as its no longer being used (image pullspecs are passed as RELATED_IMAGE_OPERATOR/RELATED_IMAGE_OPERAND/GRAPH_DATA_IMAGE env vars)

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -e
+set -ex
 
 # DEFAULT_OPERATOR_IMAGE is a placeholder for cincinnati-operator image placeholder
 # During development override this when you want to use an specific image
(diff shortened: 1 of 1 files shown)


vrutkovs:
We need https://github.com/openshift/cincinnati-operator/pull/140 to figure out why this crashes

hack/deploy.sh
```diff
@@ -10,9 +10,9 @@ DEFAULT_OPERAND_IMAGE="quay.io/cincinnati/cincinnati:latest"
 RELATED_OPERATOR_IMAGE="${RELATED_IMAGE_OPERATOR:-${DEFAULT_OPERATOR_IMAGE}}"
 RELATED_OPERAND_IMAGE="${RELATED_IMAGE_OPERAND:-${DEFAULT_OPERAND_IMAGE}}"
 
-if [ -n "$OPENSHIFT_BUILD_NAMESPACE" ]; then
-	RELATED_OPERATOR_IMAGE="registry.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:updateservice-operator"
-	GRAPH_DATA_IMAGE="registry.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:updateservice-graph-data-container"
+if [ -n "$OPENSHIFT_CI" ]; then
+	RELATED_OPERATOR_IMAGE="registry.ci.openshift.org/${NAMESPACE}/stable:updateservice-operator"
```
jottofar, line 14: Where does it get the namespace from now? Below we set `NAMESPACE` to our default where we want to create our resources but where does this CI stuff reside? I looked in the CI doc but didn't see anything related or I missed it. If it just in the same namespace as our op we can move the def up.

vrutkovs: Right, we don't need externally set `NAMESPACE` (all image pullspecs are being passed as env vars)

hack/deploy.sh
```diff
@@ -10,12 +10,8 @@ DEFAULT_OPERAND_IMAGE="quay.io/cincinnati/cincinnati:latest"
...
 RELATED_OPERAND_IMAGE="${RELATED_IMAGE_OPERAND:-${DEFAULT_OPERAND_IMAGE}}"
 
-if [ -n "$OPENSHIFT_BUILD_NAMESPACE" ]; then
-	RELATED_OPERATOR_IMAGE="registry.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:updateservice-operator"
-	GRAPH_DATA_IMAGE="registry.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:updateservice-graph-data-container"
-
-	echo "Openshift CI detected, deploying using image $RELATED_OPERATOR_IMAGE and ${GRAPH_DATA_IMAGE}"
-
+if [ -n "$OPENSHIFT_CI" ]; then
+	echo "Openshift CI detected, deploying using ${RELATED_OPERATOR_IMAGE} as operator and ${RELATED_OPERAND_IMAGE} as operand and ${GRAPH_DATA_IMAGE} as graph data image"
```
wking, line 14: do we even care about `OPENSHIFT_CI`?  Looks like we could drop this conditional, and just have:

```sh
if ! [ -n "$KUBECONFIG" ]; then
	echo "KUBECONFIG environment variable must be set."
	exit 1
fi
if ! [ -n "$GRAPH_DATA_IMAGE" ]; then
	echo "GRAPH_DATA_IMAGE environment variable must be set."
	exit 1
fi
```

checked in all cases.  Maybe with an informational:

```sh
echo "Deploying using ${RELATED_OPERATOR_IMAGE} as operator and ${RELATED_OPERAND_IMAGE} as operand and ${GRAPH_DATA_IMAGE} as graph data image"
```

for all cases?

vrutkovs: SGTM, fixed in a802f3b

PR #144, opened 2022-06-09
Bug 2094562: add support for cluster-wide proxy

PratikMahajan:


LalatenduMohanty:
I do not see code to handle the CA cert for some https proxy setting, do we plan to handle that later ?

controllers/new.go
```diff
@@ -433,6 +452,22 @@ func (k *kubeResources) newGraphDataInitContainer(instance *cv1.UpdateService) *
...
 
 func (k *kubeResources) newGraphBuilderContainer(instance *cv1.UpdateService, image string) *corev1.Container {
+
+	gbENV := []corev1.EnvVar{
+		newCMEnvVar("RUST_BACKTRACE", "gb.rust_backtrace", nameEnvConfig(instance)),
+	}
+
+	// get the cluster proxy variables, and append to ENV var if set
+	if os.Getenv("HTTP_PROXY") != ""{
+		gbENV = append(gbENV,newCMEnvVar("HTTP_PROXY", "gb.http_proxy", nameEnvConfig(instance)))
```
wking, line 462: nit: this mixing of `Getenv` and the config-map sourcing feels brittle.  Can we set the env var directly from `Getenv` for this container config instead of involving the config map?  Or can we use the config-map alone without relying on `Getenv` here (only using `Getenv` when we decide how to populate the config-map)?

PratikMahajan:
> I do not see code to handle the CA cert for some https proxy setting, do we plan to handle that later ?
Yes, we'll be handling CA certs for Https proxy settings later.

LalatenduMohanty (approved):
/lgtm

PR #146, opened 2022-07-01
Bug 2101732: Checking if an route exists with old syntax

LalatenduMohanty:
before creating the new route old as we have changed the route syntax in 8cea4f33836e0aeaf8de9cd8b1f37f569545f43f

### DIFF
--- a/controllers/updateservice_controller.go
+++ b/controllers/updateservice_controller.go
@@ -516,33 +516,32 @@ func validateRouteName(instance *cv1.UpdateService, name string, namespace strin
 }
 
 func (r *UpdateServiceReconciler) ensurePolicyEngineRoute(ctx context.Context, reqLogger logr.Logger, instance *cv1.UpdateService, resources *kubeResources) error {
-	route := resources.policyEngineRoute
-	// Set UpdateService instance as the owner and controller
-	if err := controllerutil.SetControllerReference(instance, route, r.Scheme); err != nil {
-		return err
-	}
 
-	// Check if it already exists
-	found := &routev1.Route{}
-	err := r.Client.Get(ctx, types.NamespacedName{Name: route.Name, Namespace: route.Namespace}, found)
-	if err != nil && errors.IsNotFound(err) {
+	route := resources.policyEngineRoute
+	foundRoute, err := r.findExistingRoute(ctx, reqLogger, instance, resources)
+	if err != nil {
+		handleErr(reqLogger, &instance.Status, "GetRouteFailed", err)
+		return err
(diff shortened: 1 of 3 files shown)


jottofar:
Add/modify route unit tests in controllers/updateservice_controller_test.go.

controllers/updateservice_controller.go
```diff
@@ -516,16 +516,36 @@ func validateRouteName(instance *cv1.UpdateService, name string, namespace strin
 }
 
 func (r *UpdateServiceReconciler) ensurePolicyEngineRoute(ctx context.Context, reqLogger logr.Logger, instance *cv1.UpdateService, resources *kubeResources) error {
+	oldRoute := resources.policyEngineOldRoute
 	route := resources.policyEngineRoute
+
 	// Set UpdateService instance as the owner and controller
 	if err := controllerutil.SetControllerReference(instance, route, r.Scheme); err != nil {
 		return err
 	}
```
jottofar, line 525: Shouldn't we wait to do this until we know which route, old or new, we're going to use?

controllers/updateservice_controller.go
```diff
@@ -516,16 +516,36 @@ func validateRouteName(instance *cv1.UpdateService, name string, namespace strin
...
+		return err
+	}
+
+	if uri, _, err := routeapihelpers.IngressURI(foundOldRoute, ""); err == nil {
+		instance.Status.PolicyEngineURI = uri.String()
+	} else {
+		handleErr(reqLogger, &instance.Status, "RouteIngressFailed", err)
+	}
+
+	err = r.Client.Get(ctx, types.NamespacedName{Name: route.Name, Namespace: route.Namespace}, found)
```
jottofar, line 547: Seems we should check `oldRoutePresent` and if `true` not bother with this.

LalatenduMohanty: Right. good catch. Will fix it.

controllers/updateservice_controller.go
```diff
@@ -516,16 +516,36 @@ func validateRouteName(instance *cv1.UpdateService, name string, namespace strin
 }
 
 func (r *UpdateServiceReconciler) ensurePolicyEngineRoute(ctx context.Context, reqLogger logr.Logger, instance *cv1.UpdateService, resources *kubeResources) error {
```
jottofar, line 518: Why not add a new method up front that returns the route, old or new, or nil. Then the rest of this method is fairly unchanged. if nil returned we operate on the new route but if a route is returned this method doesn't care whether it's old or new it just does it's thing (and my comments below are NA).

PR #150, opened 2022-10-13
Bug 1939855: controllers/updateservice_controller.go: handle init container changes

jottofar:
such as graphDataImage updates.

### DIFF
--- a/controllers/updateservice_controller.go
+++ b/controllers/updateservice_controller.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"regexp"
@@ -11,7 +12,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -81,7 +82,7 @@ func (r *UpdateServiceReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	instance := &cv1.UpdateService{}
 	err := r.Client.Get(ctx, req.NamespacedName, instance)
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if apiErrors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Return and don't requeue
 			return ctrl.Result{}, nil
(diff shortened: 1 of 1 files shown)


controllers/updateservice_controller.go
```diff
@@ -389,6 +391,20 @@ func (r *UpdateServiceReconciler) ensureDeployment(ctx context.Context, reqLogge
...
+	initContainers := updated.Spec.Template.Spec.InitContainers
+	for i := range initContainers {
+		var original *corev1.Container
+		switch initContainers[i].Name {
+		case "graph-data":
+			original = resources.graphDataInitContainer
+		default:
+			reqLogger.Info("encountered unexpected init container in pod", "Container.Name", initContainers[i].Name)
+			continue
+		}
```
petr-muller, line 403: Looks more wordy than I'd expect, why not simply:

```go
if initContainers[i].Name != `graph-data` {
  logger.Info
  continue
}
original = ...
initContainers[i].Image = original.Image
...
```

Is it likely that we grow more `case`s?

jottofar: I agree. I just cut-and-pasted it and been concentrating on trying to get it to test out but I'll change it.

controllers/updateservice_controller.go
```diff
@@ -389,6 +391,19 @@ func (r *UpdateServiceReconciler) ensureDeployment(ctx context.Context, reqLogge
 		containers[i].LivenessProbe = original.LivenessProbe
 		containers[i].ReadinessProbe = original.ReadinessProbe
 	}
+	initContainers := updated.Spec.Template.Spec.InitContainers
+	for i := range initContainers {
+		var original *corev1.Container
+		if initContainers[i].Name == "graph-data" {
+			original = resources.graphDataInitContainer
+		} else {
+			reqLogger.Info("encountered unexpected init container in pod", "Container.Name", initContainers[i].Name)
```
wking, line 400: why are we keeping these?  Can we stomp them off?

wking: like:

```go
updated.Spec.Template.Spec.Initcontainers = []whateverType{resources.graphDataInitContainer}
```

or some such?

wking: The Cincinnati Deployment is our operand.  We have opinions about its init containers.  Other folks might have opinions too, and attempt to inject additional init containers.  But trying to share ownership of a single Deployment is sticky, and in the CVO we've had many, many bugs as a result of initially attempting a sharing approach (openshift/cluster-version-operator#830, openshift/cluster-version-operator#804, openshift/cluster-version-operator#674, openshift/cluster-version-operator#654, openshift/cluster-version-operator#650, openshift/cluster-version-operator#563, openshift/cluster-version-operator#549 , ...), and no reports that I can remember of any users benefiting from the sharing.

PratikMahajan (commented):
/retest

PratikMahajan:

operator-e2e is pinned to a specific commit which does not contain the fix for sha-sum mismatch. 
ci/prow/operator-e2e-latest-osus  uses the latest commit from cincinnati and has passed the tests.

wking (approved):
/lgtm

jottofar:

operator-e2e is pinned to a specific commit which does not contain the fix for sha-sum mismatch.
ci/prow/operator-e2e-latest-osus uses the latest commit from cincinnati and has passed the tests.

PR #152, opened 2022-11-01
add PratikMahajan as owner and reviewer for cincinnati operator

PratikMahajan:


wking (approved):
/lgtm

PratikMahajan:
we need to override `operator-e2e` as it has a pinned commit that does not include multi-arch support. cc @jottofar

jottofar:

operator-e2e is pinned to a specific commit which does not contain the fix for sha-sum mismatch.
ci/prow/operator-e2e-latest-osus uses the latest commit from cincinnati and has passed the tests.

PR #155, opened 2023-01-10
Bumping the go version to 1.18

LalatenduMohanty:
Steps:
* edit the go.mod file
* go mod tidy
* go mod vendor

Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

PR #157, opened 2023-01-14
Update dockerfile for go118

LalatenduMohanty:
Missed this in PR https://github.com/openshift/cincinnati-operator/pull/155


wking (approved):
/lgtm

petr-muller:
Pratik says this cannot pass in https://github.com/openshift/cincinnati-operator/pull/155#pullrequestreview-1248117933 so it likely holds here as well

PR #161, opened 2023-02-09
OCPBUGS-7150: fix the mount for cluster ca certs

PratikMahajan:


controllers/new.go
```diff
@@ -637,7 +637,7 @@ func (k *kubeResources) newGraphBuilderVolumeMounts(instance *cv1.UpdateService)
 		vm = append(vm, corev1.VolumeMount{
 			Name:      NameClusterTrustedCAVolume,
 			ReadOnly:  true,
-			MountPath: "/etc/pki/ca-trust/extracted/pem/cluster-ca-bundle.crt",
+			MountPath: "/etc/pki/ca-trust/extracted/pem/ca-bundle.crt",
```
LalatenduMohanty, line 640: We still use the hard coded path here. Can we use a variable instead as we have the file name in `NameClusterCertConfigMapKey`?

LalatenduMohanty (changes requested):

PratikMahajan:
/override operator-e2e 
Using older version of cincinnati which does not have a multi-arch fix

PratikMahajan (commented):

PR #164, opened 2023-02-21
OCPBUGS-9745: add graph-data image annotation to operand pod

PratikMahajan:


controllers/names.go
```diff
@@ -50,6 +50,10 @@ func nameGraphBuilderService(instance *cv1.UpdateService) string {
 	return instance.Name + "-graph-builder"
 }
 
+func nameGraphDataPod(instance *cv1.UpdateService) string {
+	return instance.Name + "-graph-data-init-sha"
```
wking, line 54: can this be `-tag-digest` or something, to avoid being longer than the 14-character `-policy-engine` or `-graph-builder`?  I know we've had name-is-too-long issues in the past, and I'm not clear on what happens if the name returned by these helpers ends up growing long.

controllers/updateservice_controller.go
```diff
@@ -350,6 +351,38 @@ func (r *UpdateServiceReconciler) ensureTrustedClusterCA(ctx context.Context, re
 	return err
 }
 
+func (r *UpdateServiceReconciler) ensureGraphDataSHA(ctx context.Context, reqLogger logr.Logger, instance *cv1.UpdateService,
+	resources *kubeResources) (string, error){
+
+	pod := resources.graphDataPodContainer
```
wking, line 357: can we add a guard here so that we `return` quickly for by-tag graph-data pullspecs?  When the pullspec is already by-digest, launching a Pod for a tag-to-digest lookup is unnecessary overhead.

controllers/updateservice_controller.go
```diff
@@ -350,6 +351,38 @@ func (r *UpdateServiceReconciler) ensureTrustedClusterCA(ctx context.Context, re
...
+		return "", err
+	}
+
+	// Check if this pod already exists
+	found := &corev1.Pod{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, found)
+
+	if err != nil && apiErrors.IsNotFound(err) || found.Status.Phase == "Succeeded"{
+		reqLogger.Info("Creating Pod", "Namespace", pod.Namespace, "Name", pod.Name)
+		err := r.Client.Create(ctx, pod)
```
wking, line 368: Probably need a `Delete` in the `Succeeded` case.  And also maybe a delete in the "we used  to be a by-tag reference, but now the graph-data pullspec is a by-digest reference, so [we don't need to bother launching Pods anymore][1]" case?

[1]: https://github.com/openshift/cincinnati-operator/pull/164#discussion_r1119454227

controllers/updateservice_controller.go
```diff
@@ -350,6 +351,38 @@ func (r *UpdateServiceReconciler) ensureTrustedClusterCA(ctx context.Context, re
...
+	pod := resources.graphDataPodContainer
+	if err := controllerutil.SetControllerReference(instance, pod, r.Scheme); err != nil {
+		return "", err
+	}
+
+	// Check if this pod already exists
+	found := &corev1.Pod{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, found)
+
+	if err != nil && apiErrors.IsNotFound(err) || found.Status.Phase == "Succeeded"{
```
wking, line 366: `a && b || c` requires the reader to have internalized `&&` vs `||` precedence.  Can we use explicit parens, to avoid confusion?  And actually, if we're about to `create`, I'd expect:

* `err == nil`: We have a pod.  Maybe it's happy.  Maybe not.  Who cares?  Did we get a pullspec in `status`?
* `err != nil && apiErrors.IsNotFound(err)`: No pod yet.  Probably just got this UpdateService or it just got a by-tag graph-data pullspec or something.  Create the pod.
* `err != nil && !apiErrors.IsNotFound(err)`: Kube API-server is having trouble.  Bubble up to the caller.

controllers/updateservice_controller.go
```diff
@@ -350,6 +351,38 @@ func (r *UpdateServiceReconciler) ensureTrustedClusterCA(ctx context.Context, re
...
+	} else if err != nil {
+		handleErr(reqLogger, &instance.Status, "GetGraphDataPodFailed", err)
+		return "", err
+	}
+
+	if len(found.Status.ContainerStatuses) > 0 {
+		return found.Status.ContainerStatuses[0].ImageID, nil
+	}
+
+	return "", nil
```
wking, line 382: Eventually in the "we got the pod, but there's no container status yet" case, we might want to bail out and try again with a fresh pod.  Some kind of `status` check to see if the pod is in the "still trying to create containers" state or the "we've given up" state?

controllers/updateservice_controller.go
```diff
@@ -395,6 +428,12 @@ func (r *UpdateServiceReconciler) ensureDeployment(ctx context.Context, reqLogge
 		updated.Spec.Template.ObjectMeta.Annotations[key] = value
 	}
 
+	imageSHA, err := r.ensureGraphDataSHA(ctx, reqLogger, instance, resources)
+	if len(imageSHA)>0 && err == nil{
```
wking, line 432: how do we want to handle the `err != nil` case?  Can we get that bubbled up to our UpdateService `status` somehow?  Possibly by returning `err` here?

controllers/names.go
wking: Also, handling this per-instance means that  if we have multiple UpdateService sharing the same graph-data pullspec, we'll be launching redundant Pods to figure out that digest.  In practice, maybe that situation is rare enough that we don't have to worry about trying to centralize tag-to-digest lookup at the operator level where results can be shared between multiple instances.

controllers/updateservice_controller.go
PratikMahajan: added a delete on success condition

controllers/updateservice_controller.go
PratikMahajan: added

controllers/updateservice_controller.go
PratikMahajan: added

controllers/updateservice_controller.go
PratikMahajan: fixed

controllers/updateservice_controller.go
PratikMahajan: fixed

controllers/updateservice_controller.go
```diff
@@ -168,6 +170,16 @@ func (r *UpdateServiceReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	imageSHA, err := r.ensureGraphDataSHA(ctx, reqLogger, instance)
+	if err != nil{
+		reqLogger.Error(err, "ensuring GraphData SHA")
```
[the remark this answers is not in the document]
petr-muller: Also, should we return, or at least set `imageSHA` to some predictable null value? When `err!=nil` we should not rely on it having any value, but we use it later on L178

controllers/updateservice_controller.go
petr-muller: Is there a chance of racing with someone else (is it possible that we run multiple replicas?) creating the pod? It is possible that we Get, see nothing, something else creates the pod, we create and fail with conflict... I'm fine if we say that that cannot happen realistically.

controllers/updateservice_controller.go
```diff
@@ -350,8 +362,74 @@ func (r *UpdateServiceReconciler) ensureTrustedClusterCA(ctx context.Context, re
...
+		err := r.Client.Delete(ctx, pod)
+		if err != nil {
+			handleErr(reqLogger, &instance.Status, "CreateGraphDataPodFailed", err)
+		}
+		return "", err
+	}
+
+	if len(found.Status.ContainerStatuses) > 0 {
+		return found.Status.ContainerStatuses[0].ImageID, nil
+	}else{
```
petr-muller, line 423: Huh does this pass `go fmt`? I think we may be missing a `go fmt`-enforcing CI job

petr-muller: https://github.com/openshift/release/pull/36878

controllers/updateservice_controller.go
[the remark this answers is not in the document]
PratikMahajan: added a value to imageSHA where there's an error

controllers/updateservice_controller.go
wking: > because, we're creating a pod, it can only have 1 pod running at a time with the given name

Petr was pointing out that other actors (e.g. a runaway second cincinnati-operator replica) might also create a pod with the same name.  And actually, because `ensureGraphDataSHA` is checking a single UpdateService, it seems like an operator managing several by-tag UpdateService would probably struggle like:

1. Reconciling UpdateService A, create a `graph-data-tag-digest` pod.
2. Reconciling UpdateService B, pod is still running, return `"", nil`.
3. Reconciling UpdateService A, pod is still running, return `"", nil`.
4. Reconciling UpdateService B, pod is `Succeeded`, return the image ID.  But this is bad, because that image ID  was for the by-tag pullspec from UpdateService A, not the pullspec from B.

Instead, you'll want to somehow uniquify by pullspec-under test.  Perhaps `SetControllerReference` allows you to clearly associate a pod with the appropriate UpdateService.  And then Kube's built-in garbage collection will handle deleting any pods whose UpdateService is deleted.

controllers/updateservice_controller.go
```diff
@@ -186,7 +200,7 @@ func (r *UpdateServiceReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		reqLogger.Error(err, "Failed to update Status")
 	}
 
-	return ctrl.Result{}, err
+	return ctrl.Result{RequeueAfter: time.Duration(5) * time.Minute}, err
```
petr-muller, line 203: time.Duration(5) means "5 nanoseconds". did you mean simply `5 * time.Minute` ? End result is the same but with less confusion ;)

controllers/updateservice_controller.go
```diff
@@ -350,8 +364,71 @@ func (r *UpdateServiceReconciler) ensureTrustedClusterCA(ctx context.Context, re
...
+		reqLogger.Info("Creating Pod", "Namespace", pod.Namespace, "Name", pod.Name)
+		err := r.Client.Create(ctx, pod)
+		if err != nil {
+			handleErr(reqLogger, &instance.Status, "CreateGraphDataPodFailed", err)
+		}
+		return "", err
+	} else if err != nil {
+		handleErr(reqLogger, &instance.Status, "GetGraphDataPodFailed", err)
+		return "", err
+	} else if found.Status.Phase == "Succeeded" {
```
petr-muller, line 413:
```suggestion
	} else if found.Status.Phase == corev1.PodSucceeded {
```

petr-muller (approved):
/retest

PratikMahajan:
/override  ci/prow/operator-e2e-latest-osus-414 ci/prow/operator-e2e-latest-osus-414 
issues with cluster coming up, tests on other OCP versions have passed, OSUS is OCP version agnostic and it would be safe to override the tests!

PR #165, opened 2023-03-06
run go-fmt and fix formatting of all files

PratikMahajan:


api/v1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
petr-muller, line 1: qHmm... looking at this, I wonder whether we should also have a job guarding that these autogenerated files stay up to date? Shall we bump our generators sometimes? `controller-tools@v0.3.0` is ancient...

petr-muller (approved):
Changes are good - one thought inline but that's beyond the scope of a simple formatting bump, I'll take it to Slack

PR #168, opened 2023-03-15
[dist] update the dockerfile to use new osus endpoint

PratikMahajan:
use the new /graph-data endpoint on upstream update service to download the tarball for creating graph data container

### DIFF
--- a/dev/Dockerfile
+++ b/dev/Dockerfile
@@ -1,5 +1,7 @@
 FROM registry.access.redhat.com/ubi8/ubi:8.1
 
-RUN curl -L -o cincinnati-graph-data.tar.gz https://github.com/openshift/cincinnati-graph-data/archive/master.tar.gz
+RUN curl -L -o cincinnati-graph-data.tar.gz https://api.openshift.com/api/upgrades_info/graph-data
 
-CMD exec /bin/bash -c "tar xvzf cincinnati-graph-data.tar.gz -C /var/lib/cincinnati/graph-data/ --strip-components=1"
+RUN mkdir -p /var/lib/cincinnati-graph-data && tar xvzf cincinnati-graph-data.tar.gz -C /var/lib/cincinnati-graph-data/ --no-overwrite-dir
+
+CMD ["/bin/bash", "-c" ,"exec cp -rp /var/lib/cincinnati-graph-data/* /var/lib/cincinnati/graph-data"]
--- a/docs/disconnected-updateservice-operator.md
+++ b/docs/disconnected-updateservice-operator.md
@@ -105,12 +105,11 @@ In order to build a container image that contains the graph-data, which will be 
 container in the updateservice pod: 
 
 ```
-wget https://github.com/openshift/cincinnati-graph-data/archive/master.tar.gz -O graph.tar.gz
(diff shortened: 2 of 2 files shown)


dev/Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM registry.access.redhat.com/ubi8/ubi:8.1
 
-RUN curl -L -o cincinnati-graph-data.tar.gz https://github.com/openshift/cincinnati-graph-data/archive/master.tar.gz
+RUN curl -L -o cincinnati-graph-data.tar.gz https://api.openshift.com/api/upgrades_info/graph-data
 
-CMD exec /bin/bash -c "tar xvzf cincinnati-graph-data.tar.gz -C /var/lib/cincinnati/graph-data/ --strip-components=1"
+RUN mkdir -p /var/lib/cincinnati-graph-data && tar xvzf cincinnati-graph-data.tar.gz -C /var/lib/cincinnati-graph-data/ --no-overwrite-dir
+
+CMD ["/bin/bash", "-c" ,"cp -rp /var/lib/cincinnati-graph-data/* /var/lib/cincinnati/graph-data"]
```
wking, line 7: micro-nit: I'm fine with your transition to JSON-array syntax, but can we preserve the `exec` from the outgoing `CMD`?  No need to keep the shell around after it has performed the expansion and the only thing left is a `cp`.  Something like:

CMD ["/bin/bash", "-c" ,"exec cp -rp /var/lib/cincinnati-graph-data/* /var/lib/cincinnati/graph-data"]

docs/disconnected-updateservice-operator.md
```diff
@@ -108,9 +108,9 @@ container in the updateservice pod:
 wget https://github.com/openshift/cincinnati-graph-data/archive/master.tar.gz -O graph.tar.gz
```
wking, line 108: You can drop this too, if you're moving to pulling the data from inside the `Dockerfile`.

dev/Dockerfile
LalatenduMohanty: @wking You want the `exec` because we should run the command in a empty environment, is that right?

LalatenduMohanty: from `man exec`
```
exec [-cl] [-a name] [command [arguments]]
              If command is specified, it replaces the shell.  No new process is created.  The arguments become the arguments to command.  If the -l option is supplied, the shell places
              a  dash at the beginning of the zeroth argument passed to command.  This is what login(1) does.  The -c option causes command to be executed with an empty environment.  If
              -a is supplied, the shell passes name as the zeroth argument to the executed command.  If command cannot be executed for some reason, a non-interactive shell exits, unless
              the  execfail  shell option is enabled.  In that case, it returns failure.  An interactive shell returns failure if the file cannot be executed.  A subshell exits uncondi‐
              tionally if exec fails.  If command is not specified, any redirections take effect in the current shell, and the return status is 0.  If there is a redirection error,  the
              return status is 1.
```

LalatenduMohanty (approved):
/lgtm

PratikMahajan:
failing because of old cincinnati image

PR #169, opened 2023-03-30
init Dockerfile: Fix tar error setting user

jeckersb:
Since tar is run as root, it will attempt to preserve the user/group of the files in the archive by default, which leads to these errors when unpacking the graph data archive:

tar: version: Cannot change ownership to uid 1000950000, gid 0: Invalid argument

Passing the --no-same-owner flag to tar will expand the archive contents as root:root instead of trying to preserve the non-existent uid from the original archive.


PratikMahajan (approved):
/lgtm

PR #172, opened 2023-08-03
OCPBUGS-16898 : Add both trusted and additional CA to SSL_CERT_DIR

PratikMahajan:
add additionalTrustedCA as well as ClusterTrustedCA to SSL_CERT_DIR this is required when the user has both CAs set.
earlier we used to ignore additionalTrustedCA when CLusterCA was present

see: https://www.openssl.org/docs/man1.0.2/man1/c_rehash.html

### DIFF
--- a/controllers/names.go
+++ b/controllers/names.go
@@ -26,6 +26,8 @@ const (
 	namePullSecret = "pull-secret"
 	// ClusterCAMountDir is the mount path for the dir containing cluster CA
 	ClusterCAMountDir = "/etc/pki/ca-trust/extracted/cluster-ca/"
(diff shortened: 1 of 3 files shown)


petr-muller (approved):
/hold

LGTM but it seems like we're very close to being able to pre-merge test cincinnati-operator PRs so maybe we can wait several days and use it as an opportunity to test that process

LalatenduMohanty (approved):
/lgtm

petr-muller:
> LGTM but it seems like we're very close to being able to pre-merge test cincinnati-operator PRs so maybe we can wait several days and use it as an opportunity to test that process

...or not ;)

PR #175, opened 2023-09-15
OTA-859: Start using dkrv2-secondary-metadata-scrape plugin

jeckersb:


jeckersb:
This won't pass CI until it has access to a Cincinnati image which contains https://github.com/openshift/cincinnati/pull/837

PR #173, opened 2023-08-09
Include bundle manifests and Dockerfile

petr-muller:
Just ran an existing `Makefile` target /shrug:

```console
$ make bundle VERSION=5.0.2-dev
cd config/manager && /usr/bin/kustomize edit set image controller=controller:latest
/usr/bin/kustomize build config/manifests | operator-sdk generate bundle -q --overwrite --version 5.0.2-dev --channels=v1 --default-channel=v1
INFO[0000] Creating bundle.Dockerfile                   
INFO[0000] Creating bundle/metadata/annotations.yaml    
INFO[0000] Bundle metadata generated successfully       
operator-sdk bundle validate ./bundle
INFO[0000] All validation tests have completed successfully
```

The bundle will not work by itself: we are not publishing the operator images anywhere (except the productized images ofc) and we are not publishing cincinnati images anywhere suitable either (quay.io/cincinnati/cincinnati:latest is stale and quay.io/app-sre/cincinnati does not have floating or even version tags). 

So for now this work should just support building the bundles from pull request code for the purpose of testing operator and operand PRs in CI and by QE, see https://github.com/openshift/release/pull/42053

PR #196, opened 2024-07-15
cincinnati-operator: bump operator-sdk versions to v1.31.0-ocp

hongkailiu:
```console
$ podman run --platform=linux/amd64 --rm -it --entrypoint='["bash"]' registry.ci.openshift.org/origin/4.15:operator-sdk
bash-4.4$ operator-sdk version
operator-sdk version: "v1.31.0-ocp", commit: "08d08dd8e4da74c83fffa812fb8eb382eb0a072d", kubernetes version: "v1.26.0", go version: "go1.20.12 X:strictfipsruntime", GOOS: "linux", GOARCH: "amd64"
```

The [CI env.](https://github.com/openshift/release/blob/d6d69e0bf41d63f4e6848ac492498406116b4c31/ci-operator/config/openshift/cincinnati-operator/openshift-cincinnati-operator-master.yaml#L6-L9) has been moved forward and we might want to keep the version aligned with it. The manifests in bundles are [used only in CI or by QE](https://github.com/openshift/cincinnati-operator/pull/173) and thus should be safe enough to do.

The real reason that I want to bump the sdk version is that I am learning and do some practicing following the sdk doc.
Staying close with the latest version of sdk sometimes making the learning curve shorter.

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -7,7 +7,7 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=update-service-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=v1
 LABEL operators.operatorframework.io.bundle.channel.default.v1=v1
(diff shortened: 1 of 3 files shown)


petr-muller (approved):
:shipit:

PR #239, opened 2025-06-02
Operator creates cluster-trusted-ca configmap only when proxy variables are set

brunogomes011:
Hello, 

The OSUS operator is checking if proxy variables are set before deciding if the cluster-trusted-ca configmap should be created or not with the cluster trusted CA bundle. It should not checked considering using this method as proxy variables don't have any relation with cluster's certificates, but on how the applications are going to access external resources. 

In some setups, the proxy variables are not set and the registries are installed using public default CAs as root CA. In this kind of setups, the OSUS pods don't start properly because the root CAs are not imported into pod. If the public root CA is configured in image.config CR, it workarounds the issue.

Note that when no <>-trusted-ca configmap is available, the trusted CA bundle is installed on /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem. However, when image.config is configured, the certificates from configmap overwritten this configuration. If no proxies are set, no trusted bundle are present in the pod.

I would like to recommend your revision on this PR. Feel free to change or recommend anything.

### DIFF
--- a/controllers/new.go
+++ b/controllers/new.go
@@ -512,16 +512,6 @@ func (k *kubeResources) newTrustedCAConfig(instance *cv1.UpdateService, cm *core
 
 func newTrustedClusterCAConfig(instance *cv1.UpdateService, clusterCA *corev1.ConfigMap) *corev1.ConfigMap {
 
-	// check if the proxy variables are set by olm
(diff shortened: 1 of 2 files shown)


oarribas:
@wking , @PratikMahajan , could any of you check this PR for OCPBUGS-18280, or assign it correctly?

PR #264, opened 2026-06-01
OTA-1610: restrict OSUS egress to registry and proxy ports

nader-ziada:
The NetworkPolicy now only allows egress on the specific TCP ports needed: the registry port (from spec.releases, default 443) and proxy ports (from HTTP_PROXY/HTTPS_PROXY env vars). Proxy ports are excluded when the registry is cluster-internal (.svc).

following up on PR: #243 

## Summary by CodeRabbit

* **New Features**
  * NetworkPolicies now restrict UpdateService pod egress to a de-duplicated set of registry and proxy TCP ports; registry defaults to 443 when unspecified and proxy defaults (HTTPS→443, HTTP→80) are applied. Cluster-internal registries (.svc/*.svc.cluster.local) are excluded from proxy routing.

* **Bug Fixes**
  * Ensure existing NetworkPolicies are updated when egress ports change.

* **Tests**
  * Added tests for port selection, proxy handling, deduplication, cluster-internal detection, and policy generation/update.

* **Documentation**
  * Updated NetworkPolicy description to reflect restricted egress to necessary registry/proxy ports.

### DIFF
--- a/controllers/new.go
+++ b/controllers/new.go
@@ -4,8 +4,11 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
+	"net"
+	"net/url"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"text/template"
 
@@ -303,13 +306,85 @@ func (k *kubeResources) oldPolicyEngineRoute(instance *cv1.UpdateService) *route
 	}
 }
 
+func egressPorts(releases string) []int32 {
+	seen := map[int32]bool{}
+	var ports []int32
+	add := func(p int32) {
+		if !seen[p] {
+			seen[p] = true
+			ports = append(ports, p)
+		}
+	}
+
+	registry := strings.SplitN(releases, "/", 2)[0]
+	registryPort := portFromHost(registry, 443)
+	add(registryPort)
+
+	if !isClusterInternal(registry) {
+		for _, env := range []string{"HTTP_PROXY", "HTTPS_PROXY"} {
+			if v := os.Getenv(env); v != "" {
+				add(portFromProxyURL(v))
+			}
+		}
+	}
+
+	return ports
+}
+
+func portFromHost(host string, defaultPort int32) int32 {
+	_, portStr, err := net.SplitHostPort(host)
+	if err != nil {
+		return defaultPort
+	}
+	p, err := strconv.ParseInt(portStr, 10, 32)
+	if err != nil || p < 1 || p > 65535 {
+		return defaultPort
+	}
+	return int32(p)
+}
+
+func portFromProxyURL(rawURL string) int32 {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return 443
+	}
+	if p := u.Port(); p != "" {
+		n, err := strconv.ParseInt(p, 10, 32)
+		if err != nil || n < 1 || n > 65535 {
+			return 443
+		}
+		return int32(n)
+	}
+	if u.Scheme == "http" {
+		return 80
+	}
+	return 443
+}
+
+func isClusterInternal(host string) bool {
+	h, _, err := net.SplitHostPort(host)
+	if err != nil {
+		h = host
+	}
+	return strings.HasSuffix(h, ".svc") || strings.Contains(h, ".svc.")
+}
+
 func (k *kubeResources) newNetworkPolicy(instance *cv1.UpdateService) *networkingv1.NetworkPolicy {
+	ports := egressPorts(instance.Spec.Releases)
+	egressPolicyPorts := make([]networkingv1.NetworkPolicyPort, len(ports))
+	for i, p := range ports {
+		egressPolicyPorts[i] = networkingv1.NetworkPolicyPort{
+			Protocol: corev1ProtocolPtr(corev1.ProtocolTCP),
+			Port:     intOrStringPtr(intstr.FromInt32(p)),
+		}
+	}
+
 	return &networkingv1.NetworkPolicy{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      instance.Name,
 			Namespace: instance.Namespace,
 			Annotations: map[string]string{
-				DescriptionAnnotation: "This NetworkPolicy allows all egress, to support graph-builder scraping and DNS. " +
+				DescriptionAnnotation: "This NetworkPolicy restricts egress to the registry and proxy ports needed by graph-builder. " +
 					"It allows ingress from the router, to support serving policy-engine responses. " +
 					"All other ingress is blocked, including, for now, metrics scraping.",
 			},
@@ -324,7 +399,6 @@ func (k *kubeResources) newNetworkPolicy(instance *cv1.UpdateService) *networkin
 				},
 			},
 			Ingress: []networkingv1.NetworkPolicyIngressRule{{
-				// Traffic from the router to the policy-engine service
 				From: []networkingv1.NetworkPolicyPeer{{
 					NamespaceSelector: &metav1.LabelSelector{
 						MatchLabels: map[string]string{
@@ -338,12 +412,8 @@ func (k *kubeResources) newNetworkPolicy(instance *cv1.UpdateService) *networkin
 				}},
 			}},
 			Egress: []networkingv1.NetworkPolicyEgressRule{{
-				// TCP access to all ports, for registry access, possibly via proxies
-				Ports: []networkingv1.NetworkPolicyPort{{
-					Protocol: corev1ProtocolPtr(corev1.ProtocolTCP),
-				}},
+				Ports: egressPolicyPorts,
 			}, {
-				// DNS access to the cluster's openshift-dns DaemonSet.
 				To: []networkingv1.NetworkPolicyPeer{{
 					NamespaceSelector: &metav1.LabelSelector{
 						MatchLabels: map[string]string{
--- a/controllers/new.go
+++ b/controllers/new.go
@@ -366,7 +366,7 @@ func isClusterInternal(host string) bool {
 	if err != nil {
 		h = host
 	}
-	return strings.HasSuffix(h, ".svc") || strings.Contains(h, ".svc.")
+	return strings.HasSuffix(h, ".svc") || strings.HasSuffix(h, ".svc.cluster.local")
 }
 
 func (k *kubeResources) newNetworkPolicy(instance *cv1.UpdateService) *networkingv1.NetworkPolicy {
--- a/controllers/new.go
+++ b/controllers/new.go
@@ -317,15 +317,23 @@ func egressPorts(releases string) []int32 {
 	}
 
 	registry := strings.SplitN(releases, "/", 2)[0]
-	registryPort := portFromHost(registry, 443)
-	add(registryPort)
 
 	if !isClusterInternal(registry) {
 		for _, env := range []string{"HTTP_PROXY", "HTTPS_PROXY"} {
 			if v := os.Getenv(env); v != "" {
-				add(portFromProxyURL(v))
+				if p, err := portFromProxyURL(v); err == nil {
+					add(p)
+				} else {
+					log.Error(err, "Failed to parse port from the proxy environment variable", "env", env, "value", v)
+				}
 			}
 		}
+	}
+
+	// When a proxy is configured for an external registry, the pod talks to
+	// the proxy, not the registry directly, so only proxy ports are needed.
+	if len(ports) == 0 {
+		add(portFromHost(registry, 443))
 	}
 
 	return ports
@@ -343,22 +351,26 @@ func portFromHost(host string, defaultPort int32) int32 {
 	return int32(p)
 }
 
-func portFromProxyURL(rawURL string) int32 {
+func portFromProxyURL(rawURL string) (int32, error) {
 	u, err := url.Parse(rawURL)
 	if err != nil {
-		return 443
+		return 0, fmt.Errorf("failed to parse proxy URL: %w", err)
 	}
 	if p := u.Port(); p != "" {
 		n, err := strconv.ParseInt(p, 10, 32)
 		if err != nil || n < 1 || n > 65535 {
-			return 443
+			return 0, fmt.Errorf("invalid port %q in proxy URL", p)
 		}
-		return int32(n)
+		return int32(n), nil
 	}
-	if u.Scheme == "http" {
-		return 80
+	switch u.Scheme {
+	case "http":
+		return 80, nil
+	case "https":
+		return 443, nil
+	default:
+		return 0, fmt.Errorf("invalid proxy URL scheme %q", u.Scheme)
 	}
-	return 443
 }
 
 func isClusterInternal(host string) bool {
@@ -384,7 +396,7 @@ func (k *kubeResources) newNetworkPolicy(instance *cv1.UpdateService) *networkin
 			Name:      instance.Name,
 			Namespace: instance.Namespace,
 			Annotations: map[string]string{
-				DescriptionAnnotation: "This NetworkPolicy restricts egress to the registry and proxy ports needed by graph-builder. " +
+				DescriptionAnnotation: "This NetworkPolicy allows egress restricted to the necessary ports, to support graph-builder scraping and DNS. " +
 					"It allows ingress from the router, to support serving policy-engine responses. " +
 					"All other ingress is blocked, including, for now, metrics scraping.",
 			},
@@ -399,6 +411,7 @@ func (k *kubeResources) newNetworkPolicy(instance *cv1.UpdateService) *networkin
 				},
 			},
 			Ingress: []networkingv1.NetworkPolicyIngressRule{{
+				// Traffic from the router to the policy-engine service
 				From: []networkingv1.NetworkPolicyPeer{{
 					NamespaceSelector: &metav1.LabelSelector{
 						MatchLabels: map[string]string{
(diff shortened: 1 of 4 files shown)


controllers/updateservice_controller_test.go
```diff
@@ -633,6 +634,57 @@ func TestEnsurePolicyEngineRoute(t *testing.T) {
...
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: updateservice.Name, Namespace: updateservice.Namespace}, found)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Verify the port was updated to 443 in the first egress rule (registry egress)
+	assert.Equal(t, 2, len(found.Spec.Egress), "should have 2 egress rules: registry and DNS")
+	assert.Equal(t, 1, len(found.Spec.Egress[0].Ports), "first egress rule should have 1 port")
+	assert.Equal(t, intstr.FromInt32(443), *found.Spec.Egress[0].Ports[0].Port, "registry port should be 443 for quay.io")
+}
```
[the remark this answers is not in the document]
nader-ziada: fixed

controllers/new.go
```diff
@@ -324,7 +399,6 @@ func (k *kubeResources) newNetworkPolicy(instance *cv1.UpdateService) *networkin
 				},
 			},
 			Ingress: []networkingv1.NetworkPolicyIngressRule{{
-				// Traffic from the router to the policy-engine service
```
hongkailiu, line 327: Those comments really help me to remember what they are for.
Could we get them back?

Same for Line 346.

controllers/new.go
```diff
@@ -303,13 +306,85 @@ func (k *kubeResources) oldPolicyEngineRoute(instance *cv1.UpdateService) *route
...
+		}
+	}
+
 	return &networkingv1.NetworkPolicy{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      instance.Name,
 			Namespace: instance.Namespace,
 			Annotations: map[string]string{
-				DescriptionAnnotation: "This NetworkPolicy allows all egress, to support graph-builder scraping and DNS. " +
+				DescriptionAnnotation: "This NetworkPolicy restricts egress to the registry and proxy ports needed by graph-builder. " +
```
hongkailiu: I suggest not to lose information that it is still valid.

```suggestion
				DescriptionAnnotation: "This NetworkPolicy allows egress restricted to the necessary ports, to support graph-builder scraping and DNS. " +
```

A little background:

`cincinnati`: the operand of `cincinnati-operator`. It has mainly two components: graph-builder and policy-engine. graph-builder scrapes an image registry (such as `quay.io/openshift-release-dev/ocp-release` to form the update graph. 

* It requires the network access from graph-builder to the targeting registry. The registry might be public but behind a proxy or [the internal one on the cluster](https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/registry/registry-overview).
* It needs to resolve the domain name of the registry URL. Thus it requires the traffic to OpenShift DNS.

See https://github.com/openshift/cincinnati/blob/master/docs/design/cincinnati.md for more details. See https://github.com/openshift/cincinnati/blob/master/docs/design/openshift.md for the details on metadata of the release images (about what graph-builder scrapes).

controllers/new.go
```diff
@@ -338,12 +412,8 @@ func (k *kubeResources) newNetworkPolicy(instance *cv1.UpdateService) *networkin
 				}},
 			}},
 			Egress: []networkingv1.NetworkPolicyEgressRule{{
-				// TCP access to all ports, for registry access, possibly via proxies
-				Ports: []networkingv1.NetworkPolicyPort{{
-					Protocol: corev1ProtocolPtr(corev1.ProtocolTCP),
-				}},
+				Ports: egressPolicyPorts,
```
hongkailiu, line 429:
```suggestion
                // TCP access only to the necessary ports, for registry access, possibly via proxies
				Ports: egressPolicyPorts,
```

controllers/new.go
```diff
@@ -303,13 +306,85 @@ func (k *kubeResources) oldPolicyEngineRoute(instance *cv1.UpdateService) *route
...
+		return defaultPort
+	}
+	p, err := strconv.ParseInt(portStr, 10, 32)
+	if err != nil || p < 1 || p > 65535 {
+		return defaultPort
+	}
+	return int32(p)
+}
+
+func portFromProxyURL(rawURL string) int32 {
```
hongkailiu: I would bubble up the error (and then add no ports) instead of returning 443.

```suggestion
func portFromProxyURL(rawURL string) (int32, error) {
```

controllers/new_test.go
```diff
@@ -77,3 +77,146 @@ func Test_newKubeResources(t *testing.T) {
...
+		{
+			name:       "host with svc in domain name is not cluster-internal",
+			releases:   "my.svc.company.com/openshift/release",
+			httpsProxy: "http://proxy.example.com:3128",
+			expected:   []int32{443, 3128},
+		},
+		{
+			name:       "both HTTP_PROXY and HTTPS_PROXY with different ports",
+			releases:   "quay.io/openshift-release-dev/ocp-release",
+			httpProxy:  "http://proxy.example.com:8080",
```
hongkailiu: nit: maybe set up u/p there as well?

```suggestion
			httpProxy:  "http://username:password@proxy.example.com:8080",
```

controllers/new_test.go
```diff
@@ -77,3 +77,146 @@ func Test_newKubeResources(t *testing.T) {
...
+			}
+			k := &kubeResources{}
+			np := k.newNetworkPolicy(instance)
+
+			// First egress rule should have specific ports
+			assert.NotEmpty(t, np.Spec.Egress)
+			registryEgress := np.Spec.Egress[0]
+			var gotPorts []int32
+			for _, p := range registryEgress.Ports {
+				gotPorts = append(gotPorts, p.Port.IntVal)
```
hongkailiu, line 228: nit:

```suggestion
				gotPorts = append(gotPorts, p.Port.IntVal)
				assert.Equal(t, corev1.ProtocolTCP, p.Protocol)
```

controllers/updateservice_controller_test.go
```diff
@@ -633,6 +634,61 @@ func TestEnsurePolicyEngineRoute(t *testing.T) {
 	}
 }
 
+func TestEnsureNetworkPolicyUpdatesOnPortChange(t *testing.T) {
+	t.Setenv("HTTP_PROXY", "")
+	t.Setenv("HTTPS_PROXY", "")
+	t.Setenv("NO_PROXY", "")
```
hongkailiu, line 680: Are those 3 `t.Setenv()` necessary?

controllers/updateservice_controller_test.go
```diff
@@ -633,6 +634,101 @@ func TestEnsurePolicyEngineRoute(t *testing.T) {
...
+
+	assert.Equal(t, found.Spec.PodSelector.MatchLabels["app"], nameDeployment(updateservice))
+
+	assert.NotEmpty(t, found.Spec.Ingress, "should have ingress rules")
+	assert.Equal(t, intstr.FromString("policy-engine"), *found.Spec.Ingress[0].Ports[0].Port)
+
+	assert.Equal(t, 2, len(found.Spec.Egress), "should have 2 egress rules: registry and DNS")
+	assert.Equal(t, 1, len(found.Spec.Egress[0].Ports), "first egress rule should have 1 port for default registry")
+	assert.Equal(t, intstr.FromInt32(443), *found.Spec.Egress[0].Ports[0].Port, "registry port should be 443")
+	assert.Equal(t, intstr.FromInt32(5353), *found.Spec.Egress[1].Ports[0].Port, "DNS egress port should be 5353")
```
hongkailiu: nit: add the protocol in the message? Do we want to verify "egress port for UDP as well"?

```suggestion
	assert.Equal(t, intstr.FromInt32(5353), *found.Spec.Egress[1].Ports[0].Port, "DNS egress port for TCP should be 5353")
	assert.Equal(t, intstr.FromInt32(5353), *found.Spec.Egress[1].Ports[1].Port, "DNS egress port for UDP should be 5353")
```

controllers/updateservice_controller_test.go
```diff
@@ -633,6 +634,101 @@ func TestEnsurePolicyEngineRoute(t *testing.T) {
 	}
 }
 
+func TestEnsureNetworkPolicy(t *testing.T) {
+	t.Setenv("HTTP_PROXY", "")
+	t.Setenv("HTTPS_PROXY", "")
+	t.Setenv("NO_PROXY", "")
```
hongkailiu, line 640: Are those 3 `t.Setenv()` necessary?

hongkailiu (commented):
The implementation looks good to me.
Only a few minors.

controllers/updateservice_controller_test.go
nader-ziada: Yes, all 3 are necessary. The newKubeResources function (called at line 646) reads HTTP_PROXY, HTTPS_PROXY, and NO_PROXY via os.Getenv, see [new.go:658-662] If any of those are set in the test runner's environment, they'd affect the generated network policy egress rules (adding proxy ports/hosts), making the assertions at lines 671-674 non-deterministic.

The t.Setenv("") calls ensure the test runs with a clean, predictable environment regardless of what's set on the host.

controllers/updateservice_controller_test.go
nader-ziada: Yes, all 3 are necessary. The newKubeResources function (called at line 646) reads HTTP_PROXY, HTTPS_PROXY, and NO_PROXY via os.Getenv, see [new.go:658-662] If any of those are set in the test runner's environment, they'd affect the generated network policy egress rules (adding proxy ports/hosts), making the assertions at lines 671-674 non-deterministic.

The t.Setenv("") calls ensure the test runs with a clean, predictable environment regardless of what's set on the host.

controllers/new.go
nader-ziada: done

controllers/new.go
```diff
@@ -303,13 +306,87 @@ func (k *kubeResources) oldPolicyEngineRoute(instance *cv1.UpdateService) *route
...
+
+	registry := strings.SplitN(releases, "/", 2)[0]
+	registryPort := portFromHost(registry, 443)
+	add(registryPort)
+
+	if !isClusterInternal(registry) {
+		for _, env := range []string{"HTTP_PROXY", "HTTPS_PROXY"} {
+			if v := os.Getenv(env); v != "" {
+				if p, err := portFromProxyURL(v); err == nil {
+					add(p)
```
hongkailiu, line 325: I am not sure about this: When proxy is set, the egress to the registry is no longer needed (as "An HTTP proxy is a server that acts as a middleman between your device and the internet")?
It feels like it when I read https://en.wikipedia.org/wiki/Proxy_server#Web_proxy_servers

The description of OTA-1610 implies it to some extend: egress to one port only.

controllers/updateservice_controller_test.go
hongkailiu: https://pkg.go.dev/testing#T.Setenv

```
Setenv calls os.Setenv(key, value) and uses Cleanup to restore the environment variable to its original value after the test.

Because Setenv affects the whole process, it cannot be used in parallel tests or tests with parallel ancestors.
```

Does "restore" help? My feeling is that it should work as long as we do not run the tests in parallel.
Is there a way to trigger the failing run? `go test -v ./controllers/...` succeeds on my laptop without those `Setenv`

controllers/new.go
nader-ziada: Good point, updated. When a proxy is configured, the pod connects to the proxy, not the registry directly, so we now only allow egress to the proxy port(s) and skip the registry port. The registry port is still used when no proxy is set or when the registry is cluster-internal.

controllers/updateservice_controller_test.go
nader-ziada: You're right that t.Setenv doesn't work with t.Parallel(), but we don't use parallel here. The t.Setenv("") calls are to ensure the test is hermetic. If a developer has HTTP_PROXY/HTTPS_PROXY set in their shell, the generated network policy would include extra proxy ports and the assertions would fail. The "restore" cleanup is a nice bonus so we don't leak empty values into subsequent tests.

controllers/new.go
```diff
@@ -303,13 +306,91 @@ func (k *kubeResources) oldPolicyEngineRoute(instance *cv1.UpdateService) *route
...
+	}
+
+	registry := strings.SplitN(releases, "/", 2)[0]
+
+	if !isClusterInternal(registry) {
+		for _, env := range []string{"HTTP_PROXY", "HTTPS_PROXY"} {
+			if v := os.Getenv(env); v != "" {
+				if p, err := portFromProxyURL(v); err == nil {
+					add(p)
+				}
```
hongkailiu, line 328:
```suggestion
				} else {
					log.Error(err, "Failed to parse port from the proxy environment variable", "env", env, "value", v)
				}
```

controllers/new.go
```diff
@@ -303,13 +306,91 @@ func (k *kubeResources) oldPolicyEngineRoute(instance *cv1.UpdateService) *route
...
+	if p := u.Port(); p != "" {
+		n, err := strconv.ParseInt(p, 10, 32)
+		if err != nil || n < 1 || n > 65535 {
+			return 0, fmt.Errorf("invalid port in proxy URL %q", rawURL)
+		}
+		return int32(n), nil
+	}
+	if u.Scheme == "http" {
+		return 80, nil
+	}
```
hongkailiu:
```suggestion
	switch u.Scheme {
	case "http":
		return 80, nil
	case "https":
		return 443, nil
	default:
		return 0, fmt.Errorf("invalid proxy URL scheme %q", u.Scheme)
	}
```

controllers/new_test.go
```diff
@@ -77,3 +77,147 @@ func Test_newKubeResources(t *testing.T) {
...
+			httpsProxy: "http://proxy.example.com:3128",
+			expected:   []int32{3128},
+		},
+		{
+			name:       "both HTTP_PROXY and HTTPS_PROXY with different ports",
+			releases:   "quay.io/openshift-release-dev/ocp-release",
+			httpProxy:  "http://user:pass@proxy.example.com:8080",
+			httpsProxy: "http://user:pass@proxy.example.com:3128",
+			expected:   []int32{8080, 3128},
+		},
```
hongkailiu:
```suggestion
		},
		{
			name:       "bad proxy",
			releases:   "my.svc.company.com/openshift/release",
			httpsProxy: "i am a very bad proxy",
			expected:   []int32{443},
		},
		{
			name:      "bad proxy port",
			releases:  "my.svc.company.com/openshift/release",
			httpProxy: "http://user:pass@proxy.example.com:80000080",
			expected:  []int32{443},
		},
		{
			name:      "bad proxy schema",
			releases:  "my.svc.company.com/openshift/release",
			httpProxy: "bad://user:pass@proxy.example.com:80000080",
			expected:  []int32{443},
		},
```

controllers/updateservice_controller_test.go
hongkailiu: I would remove those `t.Setenv` (in other test too).

> If a developer has HTTP_PROXY/HTTPS_PROXY set in their shell

The developer has to adjust in that case. We do not need to prepare for this extreme case to keep the code as simple as possible and we have less to maintain.

But i am fine too if you want to keep them, probably not a big deal.

controllers/new.go
hongkailiu: :resolved

We need to test this to avoid overkill.
Honestly, I have never tested the Proxy feature before.

hongkailiu:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cincinnati-operator/262/pull-ci-openshift-cincinnati-operator-master-operator-e2e-old-ocp-published-graph-data/2062181951510417408

trigger ^^^ to see if it is broken everywhere.

nader-ziada:
> https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cincinnati-operator/262/pull-ci-openshift-cincinnati-operator-master-operator-e2e-old-ocp-published-graph-data/2062181951510417408
> 
> trigger ^^^ to see if it is broken everywhere.

it passed yesterday, but have been failing in the last few runs, i don't think its related to these changes

controllers/new_test.go
```diff
@@ -77,3 +77,164 @@ func Test_newKubeResources(t *testing.T) {
...
+		{
+			name:      "bad proxy port",
+			releases:  "my.svc.company.com/openshift/release",
+			httpProxy: "http://user:pass@proxy.example.com:80000080",
+			expected:  []int32{443},
+		},
+		{
+			name:      "bad proxy schema",
+			releases:  "my.svc.company.com/openshift/release",
+			httpProxy: "bad://user:pass@proxy.example.com:80000080",
```
hongkailiu: One last thing.

Sorry about my copy/paste error (the `80000080` part) from the previous suggestion.

With `80000080`, it goes to the same branch as "bad proxy port". Without it, it goes to "bad proxy".

We may remove the "bad proxy schema" case since it does not increase the coverage.

hongkailiu:
> it passed yesterday, but have been failing in the last few runs, i don't think its related to these changes

Our test step passes but it seems `must-gather` is broken somehow. We may search in slack if other devs reported the issue already.

nader-ziada:
> > it passed yesterday, but have been failing in the last few runs, i don't think its related to these changes
> 
> Our test step passes but it seems `must-gather` is broken somehow. We may search in slack if other devs reported the issue already.

nothing much in slack, but its failing for everyone: https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cincinnati-operator-master-operator-e2e-old-ocp-published-graph-data

hongkailiu:
/lgtm

/hold

In case you want to do pre-merge testing.
Free to cancel if you have other testing planned.

nader-ziada:
/hold cancel

done with manual testing and ci issue resolved

PR #266, opened 2026-06-04
NO-JIRA: Use portFromURL to replace portFromHost

hongkailiu:
The pull simplifies the code a bit and does not change any existing functions.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate egress port detection by correctly parsing proxy environment variables and external registry addresses.
  * Improved error logging that redacts sensitive URL details.
  * Better handling of explicit proxy ports and sensible fallbacks when parsing fails.

* **Tests**
  * Expanded test coverage for proxy behavior, NO_PROXY bypass rules (including CIDR/IP forms), and proxy URLs using explicit IP:port.

### DIFF
--- a/controllers/new.go
+++ b/controllers/new.go
@@ -309,7 +309,15 @@ func (k *kubeResources) oldPolicyEngineRoute(instance *cv1.UpdateService) *route
 func egressPorts(releases string) []int32 {
 	seen := map[int32]bool{}
 	var ports []int32
-	add := func(p int32) {
+	add := func(url string, addDefaultOnErr bool) {
+		p, err := portFromURL(url)
+		if err != nil {
+			log.Error(err, "Failed to parse port from url", "url", url)
+			if !addDefaultOnErr {
+				return
+			}
+			p = 443
+		}
 		if !seen[p] {
 			seen[p] = true
 			ports = append(ports, p)
@@ -321,11 +329,7 @@ func egressPorts(releases string) []int32 {
 	if !isClusterInternal(registry) {
 		for _, env := range []string{"HTTP_PROXY", "HTTPS_PROXY"} {
 			if v := os.Getenv(env); v != "" {
-				if p, err := portFromURL(v); err == nil {
-					add(p)
-				} else {
-					log.Error(err, "Failed to parse port from the proxy environment variable", "env", env, "value", v)
-				}
+				add(v, false)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


controllers/new.go
```diff
@@ -321,37 +329,20 @@ func egressPorts(releases string) []int32 {
...
+				add(v, false)
 			}
 		}
 	}
 
 	// When a proxy is configured for an external registry, the pod talks to
 	// the proxy, not the registry directly, so only proxy ports are needed.
 	if len(ports) == 0 {
-		add(portFromHost(registry, 443))
+		add("https://"+registry, true)
```
nader-ziada, line 342: The `NO_PROXY` case isn't handled by egressPorts. Before this PR it worked because the registry port was always included unconditionally, so direct connections were never blocked. Now that we only open proxy ports when a proxy is configured, if NO_PROXY includes the registry host, the pod will bypass the proxy and connect directly, but the NetworkPolicy won't have the registry port open.

controllers/new.go
```diff
@@ -321,37 +329,20 @@ func egressPorts(releases string) []int32 {
...
-	}
-	p, err := strconv.ParseInt(portStr, 10, 32)
-	if err != nil || p < 1 || p > 65535 {
-		return defaultPort
-	}
-	return int32(p)
-}
-
-func portFromProxyURL(rawURL string) (int32, error) {
+func portFromURL(rawURL string) (int32, error) {
```
nader-ziada: The function was renamed from portFromProxyURL to portFromURL to be generic, but all three error messages still hardcode "proxy URL".

controllers/new.go
```diff
@@ -309,7 +309,15 @@ func (k *kubeResources) oldPolicyEngineRoute(instance *cv1.UpdateService) *route
...
-	add := func(p int32) {
+	add := func(url string, addDefaultOnErr bool) {
+		p, err := portFromURL(url)
+		if err != nil {
+			log.Error(err, "Failed to parse port from url", "url", url)
+			if !addDefaultOnErr {
+				return
+			}
+			p = 443
+		}
```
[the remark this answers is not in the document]
hongkailiu: Done in [2742347](https://github.com/openshift/cincinnati-operator/pull/266/commits/27423470bde65435ae4f9b993f03ac65d78f1928)

controllers/new.go
hongkailiu: Done in [e7402c0](https://github.com/openshift/cincinnati-operator/pull/266/commits/e7402c041b2d6b81394be3a4fad24b1d9d7ab2c4)

controllers/new.go
hongkailiu: > Before this PR it worked because the registry port was always included unconditionally

I feel that this ^^^ might not be true.
See this example of an exiting unit test.

https://github.com/openshift/cincinnati-operator/blob/e7402c041b2d6b81394be3a4fad24b1d9d7ab2c4/controllers/new_test.go#L100-L104

`NO_PROXY` is [non-trivial in OpenShift](https://docs.redhat.com/en/documentation/openshift_container_platform/3.11/html/configuring_clusters/install-config-http-proxies).

Please review if I got it correctly by [bd444d1](https://github.com/openshift/cincinnati-operator/pull/266/commits/bd444d10326c72f0a06ef521d0789f567c070f66).

nader-ziada: yeah, your point makes sense, i haven't been working on openshift for a long as you, you got it

nader-ziada (approved):

PR #267, opened 2026-06-08
OTA-1747: add NetworkPolicy validation to e2e test

nader-ziada:
Validate that deploying an UpdateService CR produces a NetworkPolicy with correct owner references, ingress rules (router to policy-engine), and egress rules (registry port 443, DNS to openshift-dns on 5353). Also verify the policy is garbage-collected on CR deletion.

## Summary by CodeRabbit

* **Tests**
  * Added table-driven unit tests covering multiple registry-port scenarios and detailed network-policy assertions.
  * Enhanced end-to-end tests to wait for network-policy creation, validate ingress/egress rules (including registry and DNS ports/protocols and namespace/pod selectors), and confirm cleanup after resource deletion.
  * Improved test robustness by clearing proxy environment variables and strengthening polling and validation.

### DIFF
--- a/controllers/updateservice_controller_test.go
+++ b/controllers/updateservice_controller_test.go
@@ -639,39 +639,85 @@ func TestEnsureNetworkPolicy(t *testing.T) {
 	t.Setenv("HTTPS_PROXY", "")
 	t.Setenv("NO_PROXY", "")
 
-	pullSecret := newSecret()
-	updateservice := newDefaultUpdateService()
-	r := newTestReconciler(updateservice)
-
-	resources, err := newKubeResources(updateservice, testOperandImage, pullSecret, nil, nil)
-	if err != nil {
-		t.Fatal(err)
+	tests := []struct {
+		name                 string
+		existingObjs         []runtime.Object
+		expectedEgress       int
+		expectedRegistryPort int32
+	}{
+		{
+			name: "CreateNetworkPolicy",
+			existingObjs: []runtime.Object{
+				newDefaultUpdateService(),
+				newSecret(),
+			},
+			expectedEgress:       2,
+			expectedRegistryPort: 443,
+		},
+		{
+			name: "CreateNetworkPolicyCustomPort",
+			existingObjs: []runtime.Object{
+				func() *cv1.UpdateService {
+					us := newDefaultUpdateService()
+					us.Spec.Releases = "myregistry.example.com:8443/ocp-release"
+					return us
+				}(),
+				newSecret(),
+			},
+			expectedEgress:       2,
+			expectedRegistryPort: 8443,
+		},
 	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			r := newTestReconciler(test.existingObjs...)
 
-	err = r.ensureNetworkPolicy(context.TODO(), log, updateservice, resources)
-	if err != nil {
-		t.Fatal(err)
+			updateservice := &cv1.UpdateService{}
+			err := r.Client.Get(context.TODO(), types.NamespacedName{Name: testName, Namespace: testNamespace}, updateservice)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			ps, err := r.findPullSecret(context.TODO(), log, updateservice)
+			if err != nil {
+				assert.Error(t, err)
+			}
+
+			resources, err := newKubeResources(updateservice, testOperandImage, ps, nil, nil)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			err = r.ensureNetworkPolicy(context.TODO(), log, updateservice, resources)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			found := &networkingv1.NetworkPolicy{}
+			err = r.Client.Get(context.TODO(), types.NamespacedName{Name: updateservice.Name, Namespace: updateservice.Namespace}, found)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			verifyOwnerReference(t, found.ObjectMeta.OwnerReferences[0], updateservice)
+
+			// Ingress: router -> policy-engine
+			assert.Equal(t, 1, len(found.Spec.Ingress), "should have 1 ingress rule")
+			assert.Equal(t, intstr.FromString("policy-engine"), *found.Spec.Ingress[0].Ports[0].Port)
+			assert.Contains(t, found.Spec.Ingress[0].From[0].NamespaceSelector.MatchLabels, "policy-group.network.openshift.io/ingress")
+
+			// Egress: registry + DNS
+			assert.Equal(t, test.expectedEgress, len(found.Spec.Egress), "should have expected egress rules")
+			assert.Equal(t, intstr.FromInt32(test.expectedRegistryPort), *found.Spec.Egress[0].Ports[0].Port)
+			assert.Equal(t, corev1.ProtocolTCP, *found.Spec.Egress[0].Ports[0].Protocol)
+
+			// DNS egress
+			dnsEgress := found.Spec.Egress[1]
+			assert.Equal(t, "openshift-dns", dnsEgress.To[0].NamespaceSelector.MatchLabels["kubernetes.io/metadata.name"])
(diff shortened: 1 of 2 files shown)


nader-ziada:
@hongkailiu ready for review, thank you

functests/updateservice_creation_test.go
```diff
@@ -122,6 +124,107 @@ func TestCustomResource(t *testing.T) {
...
+		t.Fatalf("expected 2 DNS egress ports (TCP+UDP 5353), got %d", len(dnsEgress.Ports))
+	}
+	expectedDNSPort := intstr.FromInt32(5353)
+	for _, p := range dnsEgress.Ports {
+		if *p.Port != expectedDNSPort {
+			t.Fatalf("expected DNS port 5353, got %v", *p.Port)
+		}
+	}
+	t.Log("NetworkPolicy ingress and egress rules validated")
+
```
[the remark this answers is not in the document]
nader-ziada: fixed

controllers/updateservice_controller_test.go
```diff
@@ -639,39 +639,85 @@ func TestEnsureNetworkPolicy(t *testing.T) {
...
+			assert.Equal(t, intstr.FromInt32(test.expectedRegistryPort), *found.Spec.Egress[0].Ports[0].Port)
+			assert.Equal(t, corev1.ProtocolTCP, *found.Spec.Egress[0].Ports[0].Protocol)
+
+			// DNS egress
+			dnsEgress := found.Spec.Egress[1]
+			assert.Equal(t, "openshift-dns", dnsEgress.To[0].NamespaceSelector.MatchLabels["kubernetes.io/metadata.name"])
+			assert.Equal(t, 2, len(dnsEgress.Ports), "DNS egress should have TCP+UDP ports")
+			assert.Equal(t, intstr.FromInt32(5353), *dnsEgress.Ports[0].Port)
+		})
+	}
```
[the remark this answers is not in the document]
nader-ziada: fixed

functests/updateservice_creation_test.go
```diff
@@ -122,6 +124,119 @@ func TestCustomResource(t *testing.T) {
 	}
 	t.Logf("PodDisruptionBudget %s available", operatorName)
 
+	// Checks to see if the NetworkPolicy is available and has the expected rules.
+	var networkPolicyFound bool
+	if err := wait.Poll(retryInterval, timeout, func() (done bool, err error) {
```
hongkailiu: nit: noticed that it was used a few lines before. Maybe avoid calling the deprecated function in the new code?

```suggestion
	if err := wait.PollUntilContextTimeout(ctx, retryInterval, timeout, true, func(_ context.Context) (done bool, err error) {
```

https://github.com/openshift/cincinnati-operator/blob/10d5078ddbddc9ec154b300db3b0596e6fb97770/vendor/k8s.io/apimachinery/pkg/util/wait/poll.go#L62-L65

functests/updateservice_creation_test.go
```diff
@@ -167,8 +282,27 @@ func TestCustomResource(t *testing.T) {
 	}); err != nil {
 		t.Fatal(err)
 	}
+	if networkPolicyFound {
```
hongkailiu: Isn't `networkPolicyFound` always `true` here?
I think we can remove this the variable.

functests/updateservice_creation_test.go
```diff
@@ -122,6 +124,119 @@ func TestCustomResource(t *testing.T) {
...
+			return false, err
+		}
+		return true, nil
+	}); err != nil {
+		t.Fatal(err)
+	}
+	networkPolicyFound = true
+	t.Logf("NetworkPolicy %s available", customResourceName)
+
+	np, err := k8sClient.NetworkingV1().NetworkPolicies(operatorNamespace).Get(ctx, customResourceName, metav1.GetOptions{})
```
hongkailiu: nit: if we define `var np *networkingv1.NetworkPolicy` before `wait.Poll` we could save this `Get()`.

functests/updateservice_creation_test.go
```diff
@@ -167,8 +282,27 @@ func TestCustomResource(t *testing.T) {
 	}); err != nil {
 		t.Fatal(err)
 	}
+	if networkPolicyFound {
+		t.Log("Route reachable with NetworkPolicy enforced — policy does not break real traffic")
```
hongkailiu: This traffic is from the pod (on the CI build cluster) that runs `httpClient.Do(req)` from Line 274, to the URL of Policy engine (the ephemeral cluster).
Are you trying to verify this ingress traffic? I believe it is already verified by Line 280.

---

The following is only me throwing out ideas. No change is required to the pull.

Perhaps it is worth verifying if the unexpected traffic is blocked.
- Traffic from a pod in namespace like `default` or `kube-system` to OSUS service/pod is blocked, and
- Traffic from OSUS pod to an arbitrary server like "google.com" is blocked.

The reason we do not do ^^^: If we fail on any of the above, then that is a bug for NetWorkPolicy. We are not responsible for it.

The reason we do ^^^: cincinnati-operator is owned by us. The network policy is a part of it. We should know if it does not work as expected before our clients report bugs about it.

Let me know what you think about it.

hongkailiu (commented):
This is lovely.

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cincinnati-operator/267/pull-ci-openshift-cincinnati-operator-master-operator-e2e-new-ocp-published-graph-data/2065135222889910272/artifacts/operator-e2e-new-ocp-published-graph-data/e2e-test/build-log.txt | grep Net
    updateservice_creation_test.go:142: NetworkPolicy example available
    updateservice_creation_test.go:238: NetworkPolicy ingress and egress rules validated
    updateservice_creation_test.go:286: Route reachable with NetworkPolicy enforced — policy does not break real traffic
    updateservice_creation_test.go:307: NetworkPolicy example garbage-collected after CR deletion
```

functests/updateservice_creation_test.go
nader-ziada: The HTTP check on line 274 is the pre-existing route reachability test, not a new NetworkPolicy-specific assertion. The log line on 281 just annotates that the policy was enforced when traffic succeeded, confirming we didn't break the happy path.

Regarding negative traffic tests, I think those are interesting but better left out of this PR for a few reasons:
 - Spec assertions catch our mistakes. If we accidentally widen egress or drop the pod selector, the spec-level checks in this test will catch it. That's the intent-verification layer we own.
- Negative traffic tests verify the platform, not our operator. If a pod in default can still reach OSUS despite a correctly-declared policy, that's a CNI/NetworkPolicy controller bug. Gating our presubmits on platform behaviour adds fragility we don't control.
- Practicality. "Expected blocked" tests rely on connection timeouts, which are slow and flaky across different CI environments. They'd increase test runtime without catching bugs in our code.

That said, if we ever see reports of the policy not behaving as expected in the field, a follow-up with a dedicated negative-traffic e2e suite (outside the presubmit gate) could be valuable.

functests/updateservice_creation_test.go
nader-ziada: good catch

functests/updateservice_creation_test.go
nader-ziada: yeah, you are right

functests/updateservice_creation_test.go
nader-ziada: will remove the call to the deprecated func in all cases

functests/updateservice_creation_test.go
hongkailiu: Make sense to me about negative tests.

> The HTTP check on line 274 is the pre-existing route reachability test, not a new NetworkPolicy-specific assertion. 

The logs in the two places come and go always together: have both or have neither. That means to me they are doing the same check. If I see the log in 274, then I know already the network policy does not break the happy path.

An extra line of log is not a big deal if you want to keep it.

nader-ziada: good point, removed the extra log line

PR #268, opened 2026-06-16
OCPBUGS-60905, OCPBUGS-51196, OCPBUGS-51181: fix(metrics): metrics export through service works now

Cali0707:
The service monitor and service for metrics did not match the labels on the deployment correctly. This seems to date back to when they were first added.

## Summary by CodeRabbit

* **New Features**
  * Secured operator metrics with HTTPS on port `8443`, including TLS certificate mounting.
  * Added a dedicated metrics Service and Prometheus RBAC permissions so Prometheus can read required resources.
  * Added an integration test validating Prometheus returns operator metrics.

* **Bug Fixes**
  * Updated Prometheus `ServiceMonitor` selection to match the metrics Service using the correct label key.

* **Chores**
  * Enabled the metrics auth-proxy Service and updated the proxy role binding to use the operator ServiceAccount.
  * Changed the default metrics bind address to `:8443`.


Cali0707:
/verified by @Cali0707, running a cluster bot with this PR checked out:
```
oc port-forward svc/updateservice-operator-metrics -n install-osus-here 8443:8443
curl -sk https://localhost:8443/metrics
# HELP certwatcher_read_certificate_errors_total Total number of certificate read errors
# TYPE certwatcher_read_certificate_errors_total counter
certwatcher_read_certificate_errors_total 0
<skipping for brevity>
workqueue_work_duration_seconds_count{name="updateservice"} 0
```

Cali0707:
/verified by @Cali0707, curling the metrics svc on a cluster bot cluster

shahsahil264 (approved):
LGTM

Cali0707:
`prometheus_metrics_test.go:146: Prometheus returned 401: Unauthorized`

Looks like I need to figure out how to get auth for the Prometheus instance within this job properly...

Cali0707:
/cc @wking 

Lmk what else is needed here

controllers/updateservice_controller.go
rexagod, line 57:
```suggestion
// +kubebuilder:rbac:groups="",resources=namespaces;configmaps;pods;services;secrets,verbs=get;list;watch
```

functests/prometheus_metrics_test.go
```diff
@@ -0,0 +1,238 @@
...
+	// only watches namespaces with the openshift.io/cluster-monitoring label.
+	ns, err := k8sClient.CoreV1().Namespaces().Get(ctx, operatorNamespace, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("failed to get namespace %s: %v", operatorNamespace, err)
+	}
+	if ns.Labels == nil {
+		ns.Labels = map[string]string{}
+	}
+	if ns.Labels["openshift.io/cluster-monitoring"] != "true" {
+		ns.Labels["openshift.io/cluster-monitoring"] = "true"
```
rexagod, line 79: IIUC, this is not done by the controller in production, causing https://redhat.atlassian.net/browse/OCPBUGS-51181.

rexagod (commented):
👋🏼 Hello, I left a few comments around a bug that popped up during the current RIT rotation. If they make sense, please add the linked bug below to the PR title. I've also assigned you to the same ticket.

functests/prometheus_metrics_test.go
Cali0707: I think the issue is we were missing some operator CSV labels - my understanding is that the console looks at those to determine what should be shown: https://github.com/openshift/console/blob/6d0dcf1d76337e46d95ea1b546d5ffe692d7918d/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx#L225-L226

The console actually labels the ns itself

That said, I added the required labels into this branch 😄

pkg/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,90 @@
...
+// found, the Intermediate profile is used as default.
+func TLSConfigFromCluster(ctx context.Context, reader client.Reader) (func(*tls.Config), error) {
+	listers := &observerListers{
+		lister: &directAPIServerLister{ctx: ctx, reader: reader},
+	}
+	recorder := events.NewInMemoryRecorder(recorderComponentName)
+
+	observedConfig, errs := apiserver.ObserveTLSSecurityProfile(listers, recorder, map[string]interface{}{})
+	if len(errs) > 0 {
+		return nil, fmt.Errorf("observing TLS security profile: %v", errs[0])
```
2uasimojo: Consider `errors.Wrap()` (throughout).

2uasimojo (commented):
The TLS discovery bits seem sane to me (FWIW).

controllers/updateservice_controller.go
```diff
@@ -58,7 +58,7 @@ type UpdateServiceReconciler struct {
 // +kubebuilder:rbac:groups="apps",resources=deployments,verbs=get;list;watch
 // +kubebuilder:rbac:groups=networking.k8s.io,resources=networkpolicies,verbs=get;list;watch
 // +kubebuilder:rbac:groups="policy",resources=poddisruptionbudgets,verbs=get;list;watch
-// +kubebuilder:rbac:groups=config.openshift.io,resources=images,verbs=get;list;watch
```
wking, line 61: Don't we need this to support these calls:

```console
$ git grep -B1 'ImageConfigName'
controllers/mapper.go-// Map will return a reconcile request for a UpdateService if the event is for a
controllers/mapper.go:// ImageConfigName Image or a ConfigMap referenced by AdditionalTrustedCA.Name.
--
controllers/mapper.go-          image := &apicfgv1.Image{}
controllers/mapper.go:          err := m.client.Get(ctx, types.NamespacedName{Name: defaults.ImageConfigName, Namespace: ""}, image)
...
controllers/updateservice_controller.go-        image := &apicfgv1.Image{}
controllers/updateservice_controller.go:        err := r.Client.Get(ctx, types.NamespacedName{Name: defaults.ImageConfigName}, image)
...
vendor/github.com/openshift/cluster-image-registry-operator/pkg/defaults/defaults.go-
vendor/github.com/openshift/cluster-image-registry-operator/pkg/defaults/defaults.go:   ImageConfigName   = "cluster"
```

vendor/github.com/blang/semver/v4/json.go
```diff
@@ -0,0 +1,23 @@
+package semver
```
wking, line 1: nit for future pulls: having vendor bumps and things in the same commits as the human-generated content you added to `main.go` and similar makes it harder to review your changes (where I have to think more 😅) while excluding `vendor/` and other automated changes (where I have to think about `go.mod`, but then can trust `go mod ...` tooling to have `go.sum` and `vendor/` updated to match those `go.mod` changes, without reviewing `vendor/` line by line).

pkg/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,90 @@
...
+
+// TODO: Once openshift/api is bumped to a version that includes the Groups
+// field on TLSProfileSpec (added upstream for OCP 4.20), wire up
+// tls.Config.CurvePreferences from the profile's Groups to support
+// post-quantum key agreement (e.g. X25519MLKEM768).
+
+// TLSConfigFromCluster uses library-go's TLS security profile observer to
+// read the cluster's APIServer configuration and returns a function that
+// configures a tls.Config accordingly. If the APIServer resource is not
+// found, the Intermediate profile is used as default.
```
wking, line 42: I'm a bit nervous about how this will work vs. version skew.  The OpenShift Update Service operator is [a rolling stream operator][1], where:

> New releases are required to support all in-maintenance OpenShift Container Platform versions...

which currently means that new OSUS could be installed on [4.12 OCP clusters][2].  Checking on:

```console
api$ git diff origin/release-4.12..origin/master -- config/v1/types_tlssecurityprofile.go
```

turned up things like openshift/api#2697 removing ciphers like `DHE-RSA-AES128-GCM-SHA256` because of lacking Go support, which... I guess there's nothing we can do about that.  Other than the removal, the other change seems to be the addition of `groups` in 5.0, but [that's optional][3], so it should be fine for a new OSUS to find it unset in a 4.22 or older cluster.

[1]: https://access.redhat.com/support/policy/updates/openshift_operators#rolling-stream
[2]: https://access.redhat.com/support/policy/updates/openshift
[3]: https://github.com/openshift/api/commit/277d26dcb9#diff-2101eac4196d9b14cf061c8a6a4d40f9d8e5a77fc2690f969e7293294218afe3R212

bundle/manifests/update-service-operator.clusterserviceversion.yaml
```diff
@@ -16,12 +16,14 @@ metadata:
...
     capabilities: Basic Install
-    createdAt: "2026-06-17T20:26:03Z"
+    console.openshift.io/operator-monitoring-default: "true"
+    createdAt: "2026-07-21T18:06:48Z"
     description: Creates and maintains an OpenShift Update Service instance
     kubernetes.io/description: "This OpenShift Update Service operator Deployment
       reconciles local UpdateServices into more fundamental Kubernetes\nand OpenShift
       resources like Cincinnati Deployments and Routes, and it reports the status
       of those components in \nthe UpdateService status.\n"
+    operatorframework.io/cluster-monitoring: "true"
```
wking, line 26: [From docs][1]:

> The `operatorframework.io/cluster-monitoring=true` annotation. When this annotation is set to `true`, the OpenShift Console will update the namespace that the operator is being deployed to with the `openshift.io/cluster-monitoring=true` label. When this annotation is present, the UI will update the OpenShift Monitoring Prometheus Operator ServiceAccount with the appropriate RBAC privileges for the given namespace as well, allowing operators to be scraped by the OpenShift Monitoring Prometheus Operator.

I'm not sure if that's still current, or how it would work on clusters where [the console is not installed][2]. but it suggests that 6cade1aa8182a8b8c782fffd2cf039dd8f5f0872 might not be necessary.

[1]: https://github.com/openshift/enhancements/blob/1313bd4787e45e2a2bca50d4f058d920f7d82ceb/enhancements/olm/olm-managed-operator-metrics.md?plain=1#L123-L130
[2]: https://github.com/openshift/api/blame/aa91c5e2b221e017a0f92cf7209425bfa3b809b7/config/v1/types_cluster_version.go#L335-L337

bundle/manifests/update-service-operator.clusterserviceversion.yaml
```diff
@@ -16,12 +16,14 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2026-06-17T20:26:03Z"
+    console.openshift.io/operator-monitoring-default: "true"
```
wking, line 19: Are there docs for this one?  It sounds a lot [like `operatorframework.io/cluster-monitoring`][0], but I haven't found anything that looks like docs.  I did find [this code][1], but that also sounds a lot like the `operatorframework.io/cluster-monitoring` semantics.  Having a commit message that says "hey, reviewer and future devs, these changes are the appropiate changes to make because..." would help convince me ;)  But commit message or not, I'd like to understand what this annotation is about, and I haven't been able to turn up docs around it on my own :/

[0]: https://github.com/openshift/cincinnati-operator/pull/268#discussion_r3799529883
[1]: https://github.com/openshift/console/blob/68185f46a763099efa0c89e06074e79a53193854/frontend/packages/operator-lifecycle-manager/src/components/operator-hub/operator-hub-subscribe.tsx#L165-L167

PR #273, opened 2026-08-24
OCPBUGS-106182: chore(deps): bump golang.org/x/text to v0.39.0 (CVE-2026-56852)

shahsahil264:
Fixes CVE-2026-56852 (GHSA-jpjm-c3r5-q96r / GO-2026-5970): `norm.Iter` could enter an infinite loop when handling input containing invalid UTF-8 bytes.

**Root cause** (verified against the actual Go fix, CL 794100 / commit `5ae8e578`): invalid characters were assigned a `Properties` struct with size 0. `nextComposed` advances its input position by the current character's `Properties.size` — so on an invalid character it advanced by zero bytes, re-read the same byte, got size 0 again, and looped forever. The fix changes `compInfo` to always return `size >= 1` for invalid characters and tracks validity via a separate flag bit instead of inferring it from a zero size.

**Unlike typical transitive-dependency CVEs, this one is reachable.** `govulncheck` traces a real call path:
```
controllers/updateservice_controller.go:107 UpdateServiceReconciler.Reconcile
  -> client.statusWriter.Update
    -> norm.Form.Bytes / norm.Form.IsNormalString / norm.Form.QuickSpan / norm.Form.String
```

## Proof: govulncheck (before vs. after)

**Before (v0.7.0):**
```
Vulnerability #6: GO-2026-5970
    Infinite loop on invalid input in golang.org/x/text
  Module: golang.org/x/text
    Found in: golang.org/x/text@v0.7.0
    Fixed in: golang.org/x/text@v0.39.0
    Example traces found:
      #1: controllers/updateservice_controller.go:107:37: controllers.UpdateServiceReconciler.Reconcile calls client.statusWriter.Update, which eventually calls norm.Form.Bytes
      #2: ... norm.Form.IsNormalString
      #3: ... norm.Form.QuickSpan
      #4: ... norm.Form.String

Summary: affected by 10 vulnerabilities (reachable)
```

**After (v0.39.0):**
```
No mention of golang.org/x/text anywhere in output.
Summary: affected by 9 vulnerabilities (reachable) — down from 10.
```

## Note on scope: an unrelated pre-existing fix was required

`golang.org/x/text@v0.39.0` requires `go >= 1.25.0` as its own module minimum, so bumping this dependency also required bumping the `go` directive in `go.mod` (`1.18` -> `1.25.0`) — this is unavoidable, not incidental.

That Go version bump, in turn, caused `go vet` to start flagging two pre-existing lines in `validateRouteName` (`controllers/updateservice_controller.go`) that wrapped `fmt.Sprintf(...)` inside `fmt.Errorf(...)` instead of passing the format string directly — a redundant pattern that was previously allowed silently. Fixed as a 2-line, no-behavior-change correction; without it the build does not pass under the new Go version.

Verified via a clean baseline comparison (using a git worktree on the prior commit) that this `go vet` failure did **not** exist before this bump — it's a direct, unavoidable consequence of the version requirement, not scope creep.

## Proof: build / vet / tests

```
$ go build ./...
BUILD OK

$ go vet ./...
VET OK

$ go test ./controllers/... -v
... 56 test cases, all PASS
ok  	github.com/openshift/cincinnati-operator/controllers	0.628s

$ go test ./controllers/... -cover
ok  	github.com/openshift/cincinnati-operator/controllers	0.799s	coverage: 56.0% of statements
```

`functests` failures are pre-existing on stock `cpaas-release` too (require a live k8s cluster) — confirmed unrelated to this change via the same baseline comparison.

OCPBUGS-106182


shahsahil264:
Closing and re-targeting to `master` per feedback — these should go into the main branch for CI and future releases, not cpaas-release. Opening fresh PR against the correct base.

